PR #34, opened 2022-06-06
Bundle Saturn L2 node

### LINKED ISSUE #19: Bundled Saturn L2 Node
- Bundle Saturn L2 Node binary into the Station app - see https://github.com/filecoin-project/saturn-l2
- Create a semi-automated workflow for pulling new versions, e.g. “run this script, commit these changed files, open a pull request”.

Dependencies:
- [x] We need to setup build & release process for saturn-l2 - see https://github.com/filecoin-project/saturn-l2/issues/13

bajtos:
- Add a build step to download Saturn L2 Node binaries from project's [GitHub Releases](https://github.com/filecoin-project/saturn-l2/releases)
- Modify Electron Builder configuration to include Saturn L2 binary in packaged Resources
- Add a dummy main script to print the path of the Saturn L2 Node binary and verify that the path exists

Close #19

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,8 @@ jobs:
 
       - name: Build
         run: npm run build
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Test
         run: npm run test
@@ -95,9 +97,6 @@ jobs:
       - name: Install dependencies
         run: npm ci --prefer-offline --no-audit --progress=false --cache ${{ github.workspace }}/.cache/npm
 
-      - name: Build
-        run: npm run build
-
       - name: Get tag
         id: tag
         if: startsWith(github.ref, 'refs/tags/v')
@@ -123,6 +122,7 @@ jobs:
           # Apple notarization
           APPLEID: ${{ secrets.apple_id }}
           APPLEIDPASS: ${{ secrets.apple_id_pass }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Show dist/
         run: du -sh dist/ && ls -l dist/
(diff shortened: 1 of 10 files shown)


bajtos:
Thank you, @lidel! Your insights are always helpful 👏🏻 

I have two more questions if you don't mind. 

(1) In the current implementation, I am downloading release archives for all combinations of (platform, architecture). This would be useful if we were cross-compiling for multiple platforms from the same source tree. However, IIUC, our build setup is always building for the current platform & architecture only. That means we can download only one release archive. Is it a reasonable assumption?

(2) Do you happen to know what's the story for building Electron for M1 Macs?
- I am working on a darwin-arm64 machine, building darwin-arm64 package works well locally.
- The next step is to build darwin-arm64 as part of our GitHub Actions release workflow. I believe GHA does not offer darwin-arm64 runners yet, one has to setup a self-hosted runner. But then do we need that? Isn't it possible to build darwin-arm64 package on an x64 host?
- Finally, there are two options how to package macOS applications - either provide two installers, one for x64 another for arm64, or create a universal app bundling both flavours into a single binary. To do that, we need to build both x64 & arm64 versions from the same source tree, which brings me back to the first question whether it's ok to download only one release archive of saturn-l2.

Support for darwin-arm64 is most likely out of scope of MVP, I created a new GH issue for that: https://github.com/filecoin-project/filecoin-station/issues/35

bajtos:
In [8b9208e](https://github.com/filecoin-project/filecoin-station/pull/34/commits/8b9208ea4c9a06506ce13629b44699e08f02d9c8), I reworked the implementation to use GITHUB_TOKEN when calling GitHub and fetch release archives for the current platform only.

Using the dist files attached to GHA run, I verified on the following platforms that Station contains saturn-l2 binary that can be executed: MacOS M1, Windows x64.

PR #38, opened 2022-06-13
docs: add more content to Developer Guide

bajtos:
Connect to #22

Issue #22, opened 2022-05-25
Minimal Developer Guide


bajtos:
- architecture overview
- how to set up the dev env
- how to make changes, how to run the tests

PR #39, opened 2022-06-15
build: move Saturn L2 download to `postinstall`

bajtos:
- Rename the script from `build:saturn` to `install-saturn`
- Run this script from `postinstall` hook

This is a follow-up for #38, see also #22

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,9 +48,6 @@ jobs:
       - name: Build
         run: npm run build
 
-      - name: Test
-        run: npm run test
-
       - name: Test end-to-end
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -48,9 +48,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Test
-        run: npm run test
-
```
bajtos, line 53: Note: `npm run test` is running `build` + `test:e2e`. Both of these steps are already executed by this workflow.

Issue #20, opened 2022-05-25
Start/stop Saturn L2 Node + show status


bajtos:
Add UI elements to start and stop the Saturn L2 Node, show node status in the UI. 

**Notes**
- Include minimal unit/integration/e2e tests. 
- Use a hard-coded dummy FIL address from a test-net account.
- WIP: https://github.com/filecoin-project/filecoin-station/tree/manage-node

**Dependencies**
- #18 
- #19

bajtos:
Done 🎉

PR #40, opened 2022-06-15
Start/stop Saturn L2 node

bajtos:
Add functions to start and stop the saturn node, contextBridge functions to access the node status, and a toggle button to saturn page as the temporary UI.

Connect to #20


renderer/src/Saturn.tsx
```diff
@@ -0,0 +1,69 @@
...
+          uncheckedIcon={
+            <div style={textStyle}>Off</div>
+          }
+        />
+      </label>
+    </div>
+  )
+}
+
+async function isSaturnNodeRunning (): Promise<boolean> {
```
guanzo, line 59: Do you prefer explicit typing even when the compiler can infer it?

guanzo:
Left some comments, otherwise lgtm

guanzo:
Left some comments, otherwise lgtm

PR #60, opened 2022-07-26
ADR: Module Configuration

bajtos:
- chore: enable Prettier for Markdown files
- build: disable git's autocrlf on Windows CI
- docs: template for Architecture Decision Records
- docs: add Module Configuration ADR

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,9 @@
+# We are using https://standardjs.com coding style. It is unfortunately incompatible with Prettier;
+# see https://github.com/standard/standard/issues/996 and the linked issues.
+# Standard does not handle Markdown files though, so we want to use Prettier formatting.
+# The goal of the configuration below is to configure Prettier to lint only Markdown files.
+**/*
+!**/*.md
+
+# Let's keep LICENSE.md in the same formatting as we use in other PL repositories
+LICENSE.md
--- a/docs/architecture-decision-records/2022-07-module-configuration.md
+++ b/docs/architecture-decision-records/2022-07-module-configuration.md
@@ -1,0 +1,100 @@
+# Station Module Configuration
+
+- Status: ACCEPTED
+
+## Context
+
+Station modules want to allow Station users to tweak certain aspects of their
+runtime behaviour.
+
+For example, Saturn L2 Node lets users control how much disk space they are
+willing to offer for the cache.
+
+This configuration is module-specific. Each module has different options.
+
+The Station communicates the configuration to the Module via environment
+variables passed to the module backend process.
+
+## Options Considered
+
+### (1) Each module provides config WebUI
+
+The initial idea was to let each Station Module implement its configuration
+screen as part of the module WebUI and persist the configuration in a file. The
+Station would provide the directory where to keep the config file.
+
+We quickly realised this was not optimal. The Saturn module is written in Go and
+expects users to provide all configuration options via environment variables in
+the spirit of [The Twelve-Factor App](https://12factor.net). Implementing a
+configuration layer with GUI requires a lot of work inside Saturn L2 that's not
+adding much value to Saturn:
+
+1. A REST API to read & write the configuration from the WebUI.
+2. Code to persist the configuration in a file. Also, handle possible conflicts
+   between the config file and environment variables.
+3. The config editor GUI.
+4. Support for changing configuration options at runtime with zero downtime.
+
+### (2) Module provides config schema; Station provides configuration WebUI
+
+A different approach is to mimic what VisualStudio Code offers to extensions:
+
+- Modules provide schema describing their configuration, and
+- Station builds a configuration GUI based on the module-provided schema.
+
+This creates a value proposition offered by the Station to future Module
+authors: modules don't have to implement a user-friendly configuration GUI; the
+Station will provide it for free.
+
+There are some caveats, though. Supporting complex configuration formats is a
+lot of work. VS Code has been evolving the configuration editor for years, and
+there are still edge cases where the user has to edit the configuration in the
+JSON file.
+
+Some edge cases to consider:
+
+- A config option is an array of primitive values (an array of strings, an array
+  of numbers)
+- A config option is an enum - there is a set of valid values to pick from.
+- Combine 1+2, and we have an array of enum values.
+- A config option is an object (a key-value map). In some cases, sub-options are
+  well defined - we know what the valid keys are and what value types each key
+  accepts. In other cases, this can be entirely dynamic.
+
+## Decision
+
+We have decided to implement the second option, where the Station provides the
+configuration GUI for Modules.
+
+We will start by supporting:
+
+- string keys mapped directly to environment variables
+- string or integer values
+- optional and required config options
+- string descriptions of config options
+
+We will implement support for more complex configurations later.
+
+To keep our config schema easy to extend and forward-compatible, it should be a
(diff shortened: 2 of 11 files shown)


docs/architecture-decision-records/2022-07-module-configuration.md
```diff
@@ -0,0 +1,95 @@
...
+- string keys mapped directly to environment variables
+- string or integer values
+- optional and required config options
+
+We will implement support for more complex configurations later.
+
+To keep our config schema easy to extend and forward-compatible, it should be a
+superset of
+[JSON Schema](https://json-schema.org/understanding-json-schema/reference/index.html).
+
```
bajtos, line 80: A discussion point: how will a Module provide the config schema?

I like the current module distribution model, where each module is a single binary file.

I propose requiring modules to provide a special CLI arg (e.g. `--config-schema`) for obtaining the configuration schema.

1. Saturn L2 can keep the config schema file alongside the Go source code.
2. Saturn L2 will use Go embed to bundle the config schema.
3. Saturn L2 detects `--config-schema` arg and prints config schema to stdout.
4. Station runs `saturn-l2 --config-schema` to get its config schema. We can do this as part of our build step. In the future, this can happen at module install time.

Thoughts?

Maybe this should go to a different ADR file?

docs/architecture-decision-records/2022-07-module-configuration.md
```diff
@@ -0,0 +1,95 @@
...
+We have decided to implement the second option, where the Station provides the
+configuration GUI for Modules.
+
+We will start by supporting:
+
+- string keys mapped directly to environment variables
+- string or integer values
+- optional and required config options
+
+We will implement support for more complex configurations later.
```
guanzo, line 75: If there's a robust open source lib that solves our future problems now, it's worth considering starting with that. I'll take some time to look around and report back. 

I like starting simple but like you said wrt vscode config, this could get out of hand in a hurry, and if someone has already solved these problems then we should piggy back off them instead of reinventing the wheel

docs/architecture-decision-records/2022-07-module-configuration.md
guanzo: sgtm. How should the station/module handle breaking changes or updates? Say it wants to rename a config key, or add a config option, or update a default value?

We can diff the new config with the existing config and handle the various cases.

New key: Add to existing config
Missing key: Remove from existing config
Updated value: Update existing value except if the user set a value?

As for how to differentiate a renamed key vs a new key, I'm not sure. 
* Maybe it's easier to not allow it but we can't enforce it.
* Module author can "rename" by removing old key and adding a new key.
  *  But we need to retain the user specified value if it exists.

docs/architecture-decision-records/2022-07-module-configuration.md
bajtos: 💯

I agree with you, I wish there was a ready-made React component we could use. I did a quick search and did not find any 😞 Maybe you will have more luck 🤞🏻

docs/architecture-decision-records/2022-07-module-configuration.md
bajtos: > Say it wants to rename a config key, add a config option, or update a default value?

This is the classic API evolution and versioning problem, isn't it?

VS Code has the concept of deprecating config options. I guess the process goes like this:
1. Add a new option & deprecate the old one. Internally, use the deprecated option as a fallback if the new option is not provided.
2. Eventually, after some time passed, release a major version that removes the deprecated option.

I can imagine that the config schema can have a section describing migration paths from previous versions, akin to database migrations. Migration entries can be instructions like "delete this config field", "rename field A to B", and so on.

Anyhow, I think these complications are out of the scope of the initial release. I'll capture them in the ADR, though.

docs/architecture-decision-records/2022-07-module-configuration.md
bajtos: Asked on Twitter: https://twitter.com/bajtos/status/1552194253471744000

Responses so far:

- https://github.com/rjsf-team/react-jsonschema-form
- [nx](https://nx.dev) has that for its plugins. (It's not clear to me where to find the implementation, though.)

bajtos: [react-jsonschema-form](https://github.com/rjsf-team/react-jsonschema-form) looks sweet!

I see one issue, though - they don't support TailwindCSS/daisyUI, which we want to use for visual styling.

.prettierignore
```diff
@@ -0,0 +1,11 @@
...
+assets
+build
+main
+node_modules
+renderer
+test
+*.js
+*.json
+**/*.yml
+**/*.yaml
```
juliangruber, line 11: Why exclude these files?

docs/architecture-decision-records/2022-07-module-configuration.md
```diff
@@ -0,0 +1,99 @@
...
+## Decision
+
+We have decided to implement the second option, where the Station provides the
+configuration GUI for Modules.
+
+We will start by supporting:
+
+- string keys mapped directly to environment variables
+- string or integer values
+- optional and required config options
```
juliangruber, line 73:
```suggestion
- optional and required config options
- string descriptions of config options
```

docs/architecture-decision-records/2022-07-module-configuration.md
juliangruber: This is adding complexity that we can address once we need it, compared to mapping key/value pairs directly to html form elements, I think? My only concern was for the schema to be future proof, but I think also there we're good as we're working closely enough with the bleeding-edge modules that this is not an issue yet.

docs/architecture-decision-records/2022-07-module-configuration.md
juliangruber: I agree that this is outside the scope for now, but very good that you brought up these upgrade paths for the future.

juliangruber (changes requested):

.prettierignore
bajtos: We are using [standard](https://www.npmjs.com/package/standard) coding style. It is unfortunately incompatible with Prettier; see https://github.com/standard/standard/issues/996 and the linked issues.

My goal here is to configure Prettier to lint only Markdown files.

I am ok with adopting Prettier for Yaml & JSON files if you like?

Please note that `npm` may produce `package.json` in a different format than Prettier, so we should probably keep that one excluded.

.prettierignore
```diff
@@ -0,0 +1,11 @@
+LICENSE.md
```
bajtos, line 1: I am excluding this file to avoid whitespace-only changes in the license text and to keep the file consistent with what we have in other repositories.

.prettierignore
juliangruber:
```suggestion
*
!**.md
```

> My goal here is to configure Prettier to lint only Markdown files.

I think this expresses the intend more clearly (assuming it works).

> I am ok with adopting Prettier for Yaml & JSON files if you like?

I think standard covers these as well, no need from me.

bajtos: Great suggestion!

I reworked the prettier ignore file based on your suggestion and also added comments to explain what's going on.

See 30eea22

@juliangruber LGTY now?

juliangruber (approved):
sounds like a plan!

Issue #53, opened 2022-07-14
Improve UX for auto-upgrading


bajtos:
After a new version was automatically downloaded in the background:

- [x] Add a menu item to trigger an explicit & immediate check for a new version
  → #62
- [x] Change the Tray icon to notify the user about a pending upgrade. An example from Docker Desktop: 
   <img width="38" alt="Screenshot 2022-08-09 at 15 42 30" src="https://user-images.githubusercontent.com/1140553/183700918-9d960f37-cc39-4173-a220-c1888de8b43d.png"> <img width="42" alt="Screenshot 2022-08-09 at 15 45 00" src="https://user-images.githubusercontent.com/1140553/183700925-60c47eaa-46a0-4d86-95f6-badecd0991d2.png">
  → https://github.com/filecoin-project/filecoin-station/issues/166
- [x] Remove OS-level notifications about a new version being available

Out of scope:

- Show a banner at the top of the window to let the user know about the new version and ask them to restart the app to upgrade. Or a button in the top window bar. 
  → https://github.com/filecoin-project/filecoin-station/issues/137
- Auto-restart app the app when it's running in the tray (but not when it's open)

bajtos:
See https://github.com/ipfs/ipfs-desktop/pull/841/files and the latest version in https://github.com/ipfs/ipfs-desktop/blob/main/src/auto-updater/index.js

juliangruber:
> * After a new version was automatically downloaded in background, show a banner at the top of the window to let the user know about the new version and ask them to restart the app to upgrade.

We could also use desktop notifications for this, so that we can iterate without UI needs. This will also show it if the app is minimized / hidden in the menu bar.

If we want people to always update timely, then I think adding this proposed banner, and even changing the application icon in the menu bar (add red exclamation mark or so) will help enforce an upgrade.

If we want to be even more enforcing, we could automatically restart the app for them. If they have the window currently visible, we could show a dialog "will be restarted in ...". I wonder how important timely updates are for us?

bajtos:
I prefer to avoid obtrusive notifications as much as possible. IMO, we should gently nudge our users to update soon, but not push them.

I like the idea of changing the application icon (the tray icon) 👍🏻

bajtos:
> I prefer to avoid obtrusive notifications as much as possible. IMO, we should gently nudge our users to update soon, but not push them.

FWIW, the current implementation is already showing OS-level notification 😞 
[...]

PR #62, opened 2022-07-27
feat: tray & app menu item "Check for Updates..."

bajtos:
Add "Check for Updates..." to the Tray menu. On macOS, add this item to the Application menu too. While the check is running, replace "Check for Updates" with a disabled item "Checking for Updates".

See https://github.com/filecoin-project/filecoin-station/issues/53

The implementation is heavily inspired by IPFS Desktop:
- [src/auto-updater/index.js](https://github.com/ipfs/ipfs-desktop/blob/b269cde5231915d2930eefba55b706012843a155/src/auto-updater/index.js)
- [src/tray.js](https://github.com/ipfs/ipfs-desktop/blob/b269cde5231915d2930eefba55b706012843a155/src/tray.js)

### Out of scope

- Reworking the OS-level notification
- Showing a banner at the top of the web UI when an update is ready

### Screenshots

A new macOS application menu item:

Screenshot 2022-07-27 at 14 48 56

While checking for updates:

Screenshot 2022-07-27 at 14 49 06

A new Tray menu item:

Screenshot 2022-07-27 at 14 50 35

While checking for updates:

Screenshot 2022-07-27 at 14 50 41

Various dialogs shown by the app:

Screenshot 2022-07-27 at 14 49 33

Screenshot 2022-07-27 at 14 51 43
Screenshot 2022-07-27 at 14 51 56
Screenshot 2022-07-27 at 14 52 05

The annoying notification we should remove later:

Screenshot 2022-07-27 at 14 51 24

### DIFF
--- a/main/app-menu.js
+++ b/main/app-menu.js
@@ -1,0 +1,58 @@
+const { Menu, MenuItem, ipcMain } = require('electron')
+const { ipcMainEvents } = require('./ipc')
+
+function setupAppMenu (/** @type {import('./typings').Context} */ ctx) {
+  // Add "Check for updates..." item to the Application menu on MacOS
+  // GitHub issues tracking the work to add this menu item on other platforms:
+  //  - Windows: https://github.com/filecoin-project/filecoin-station/issues/63
+  //  - Linux: https://github.com/filecoin-project/filecoin-station/issues/64
+  if (process.platform !== 'darwin') return
+
+  const menu = Menu.getApplicationMenu()
+  if (!menu) return
+
+  menu.items[0].submenu?.insert(1, new MenuItem({
+    id: 'checkForUpdates',
+    label: 'Check For Updates...',
+    click: () => { ctx.manualCheckForUpdates() }
+  }))
+  menu.items[0].submenu?.insert(2, new MenuItem({
+    id: 'checkingForUpdates',
+    label: 'Checking For Updates',
+    enabled: false,
+    visible: false
+  }))
+
+  Menu.setApplicationMenu(menu)
+  setupIpcEventListeners(menu)
+}
+
+/**
+ * @param {Electron.Menu} appMenu
+ */
+function setupIpcEventListeners (appMenu) {
+  ipcMain.on(ipcMainEvents.UPDATE_CHECK_STARTED, () => {
+    getItemById('checkForUpdates').visible = false
+    getItemById('checkingForUpdates').visible = true
+  })
+
+  ipcMain.on(ipcMainEvents.UPDATE_CHECK_FINISHED, () => {
+    getItemById('checkForUpdates').visible = true
+    getItemById('checkingForUpdates').visible = false
+  })
+
+  /**
+   * Get an item from the main app menu or fail with a useful error message.
+   * @param {string} id
+   * @returns {Electron.MenuItem}
+   */
+  function getItemById (id) {
+    const item = appMenu.getMenuItemById(id)
+    if (!item) throw new Error(`Unknown app menu item id: ${id}`)
+    return item
+  }
+}
+
+module.exports = {
+  setupAppMenu
+}
--- a/main/updater.js
+++ b/main/updater.js
@@ -1,55 +1,30 @@
-const { BrowserWindow, app, dialog, Notification } = require('electron')
+const { BrowserWindow, app, Notification, shell } = require('electron')
 const { autoUpdater } = require('electron-updater')
-const fs = require('fs')
+const { ipcMain } = require('electron/main')
+const ms = require('ms')
 const log = require('electron-log').scope('updater')
+
+const { showDialogSync } = require('./dialog')
+const { ipcMainEvents } = require('./ipc')
 
 // must be global to avoid gc
 let updateNotification = null
 
+let checkingManually = false
+
+function beforeQuitCleanup () {
+  BrowserWindow.getAllWindows().forEach(w => w.removeAllListeners('close'))
+  app.removeAllListeners('window-all-closed')
+}
+
 function setup (/** @type {import('./typings').Context} */ _ctx) {
   autoUpdater.autoDownload = false // we download manually in 'update-available'
 
-  autoUpdater.on('error', err => log.error('error', err))
-  autoUpdater.on('update-available', async (/* { version, releaseNotes } */) => {
-    try {
-      log.info('update-available, downloading..')
-      await autoUpdater.downloadUpdate()
-      log.info('update-available, downloaded')
-    } catch (err) {
-      log.error('update-available error', err)
-    }
-  })
-  autoUpdater.on('update-not-available', (/* { version } */) => {
-    log.info('update not available')
-  })
-  autoUpdater.on('update-downloaded', ({ version }) => {
-    log.info(`update to ${version} downloaded`)
-    const showUpdateDialog = () => {
-      const opt = dialog.showMessageBoxSync({
-        title: 'Update Filecoin Station',
-        message: `An update to Filecoin Station ${version} is available. Would you like to install it now?`,
-        type: 'info',
-        buttons: ['Later', 'Install now']
-      })
-      if (opt === 1) { // install now
-        setImmediate(async () => {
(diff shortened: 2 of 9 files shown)


main/app-menu.js
```diff
@@ -0,0 +1,56 @@
...
+    id: 'checkForUpdates',
+    label: 'Check For Updates...',
+    click: () => { ctx.manualCheckForUpdates() }
+  }))
+  menu.items[0].submenu?.insert(2, new MenuItem({
+    id: 'checkingForUpdates',
+    label: 'Checking For Updates',
+    enabled: false,
+    visible: false
+  }))
```
bajtos, line 21: I am not sure where other Electron applications place "Check for Updates..." item on Windows & Linux, so I am adding this feature to macOS only to keep things simple.

If you think we should add this item to the application menu on other platforms too, then let's create follow-up GH issues, one per platform.

main/updater.js
```diff
@@ -1,69 +1,158 @@
...
+    checkingManually = true
+    checkForUpdatesInBackground()
+  }
+}
+
+function checkForUpdatesInBackground () {
+  ipcMain.emit(ipcMainEvents.UPDATE_CHECK_STARTED)
+
+  // TODO: replace this with autoUpdater.checkForUpdatesAndNotify()
+  autoUpdater.checkForUpdates()
```
bajtos, line 60: This todo comment is coming from the original source code. I am not sure if it's still relevant. Do you remember @lidel why you would like to switch from `autoUpdater.checkForUpdates()` to `autoUpdater.checkForUpdatesAndNotify()`?

main/app-menu.js
```diff
@@ -0,0 +1,56 @@
...
+  Menu.setApplicationMenu(menu)
+  setupIpcEventListeners(menu)
+}
+
+/**
+ * @param {Electron.Menu} appMenu
+ */
+function setupIpcEventListeners (appMenu) {
+  ipcMain.on(ipcMainEvents.UPDATE_CHECK_STARTED, () => {
+    getItemById('checkForUpdates').enabled = false
```
juliangruber, line 32:
```suggestion
```

I don't see the purpose of this, afaic if an item is invisible it's also disabled

main/app-menu.js
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * @param {string} id
+   * @returns {Electron.MenuItem}
+   */
+  function getItemById (id) {
+    const item = appMenu.getMenuItemById(id)
+    if (!item) throw new Error(`Unknown app menu item id: ${id}`)
+    return item
+  }
```
juliangruber, line 51: Could we also do without this function? We'd get an error in the consuming lines anyway.

main/app-menu.js
```diff
@@ -0,0 +1,56 @@
+const { Menu, MenuItem, ipcMain } = require('electron')
+const { ipcMainEvents } = require('./ipc')
+
+function setupAppMenu (/** @type {import('./typings').Context} */ ctx) {
+  if (process.platform !== 'darwin') return
+
+  // Add "Check for updates..." item to the Application menu on MacOS
```
juliangruber, line 7: Did you consider Windows for this PR as well? If we don't add it now then an issue will help keeping the implementations in sync in the future

main/app-menu.js
juliangruber: Same question applies to the tray configuration

main/app-menu.js
juliangruber: Same question applies to the tray configuration

main/updater.js
```diff
@@ -1,69 +1,158 @@
...
+        buttons: ['Close']
+      })
     }
+    return
+  }
+
+  setup(ctx)
+
+  checkForUpdatesInBackground() // async check on startup
+  setInterval(checkForUpdatesInBackground, 43200000) // every 12 hours
```
juliangruber, line 47:
```suggestion
  setInterval(checkForUpdatesInBackground, ms('12h'))
```

https://github.com/vercel/ms helps clean up some comments

juliangruber (changes requested):

main/app-menu.js
bajtos: To be honest, I am not sure about this either. I have simply copied what I found in ipfs-desktop:

https://github.com/ipfs/ipfs-desktop/blob/b269cde5231915d2930eefba55b706012843a155/src/tray.js#L305-L307

```js
    menu.getMenuItemById('checkForUpdates').enabled = !isUpdating
    menu.getMenuItemById('checkForUpdates').visible = !isUpdating
    menu.getMenuItemById('checkingForUpdates').visible = isUpdating
```

main/updater.js
bajtos: Personally, I am cautious about adding new dependencies to the project. Is the maintenance cost of keeping `ms` up to date worth the benefit of avoiding a comment?

BTW my PR is not changing this part, I just moved existing code & the comment.

main/app-menu.js
bajtos: We are thinking alike 😀

https://github.com/filecoin-project/filecoin-station/pull/62#discussion_r931039942

main/app-menu.js
juliangruber: I think we should add these items there too 👍 

https://github.com/filecoin-project/filecoin-station/issues/63
https://github.com/filecoin-project/filecoin-station/issues/64

main/updater.js
juliangruber: To me this is a DX improvement that overweighs the maintenance aspect, especially once we use it in more places. In my experience of using `ms` it hasn't been a maintenance burden.

main/app-menu.js
bajtos: Thanks for creating the GH issues, @juliangruber 🙇🏻 I'll mention them in a code comment.

main/app-menu.js
bajtos: This function serves two purposes:
- Make the TypeScript compiler happy
- Report a more helpful error message than _Uncaught TypeError: Cannot set properties of undefined (setting 'visible')`_

bajtos:
@juliangruber I addressed your feedback in [172bf3b](https://github.com/filecoin-project/filecoin-station/pull/62/commits/172bf3bd193196a1d203c7d20a475bc5b783a509).

- Added a code comment about Windows & Linux
- Reworked update check interval `ms('12h')` 

LGTY now?

main/app-menu.js
juliangruber: Is this even run through ts, given the .js extension?

main/app-menu.js
```diff
@@ -0,0 +1,57 @@
...
+    getItemById('checkForUpdates').visible = false
+    getItemById('checkingForUpdates').visible = true
+  })
+
+  ipcMain.on(ipcMainEvents.UPDATE_CHECK_FINISHED, () => {
+    getItemById('checkForUpdates').visible = true
+    getItemById('checkingForUpdates').visible = false
+  })
+
+  /**
```
juliangruber, line 44:
```suggestion
  /**
   * Get item or fail with a useful error message.
```

main/app-menu.js
bajtos: Yes, we check types in JS files, that's why there are JSDoc comments in most places.

https://github.com/filecoin-project/filecoin-station/blob/62b5de84631c0bd3b01aa70407f4f1ccdf6ca021/tsconfig.json#L7-L8

I found this setup useful so far. I have full auto-completion support in VS Code and the type checker can catch many programming errors.

main/app-menu.js
bajtos: Done in [af82612](https://github.com/filecoin-project/filecoin-station/pull/62/commits/af8261296e09e04876d64135d36588b2360d2e52)

juliangruber (approved):

main/updater.js
lidel: I vaguely remember that the rationale was that it is less code to maintain, and that meant smaller surface for breaking autoupdates when upstream APIs change in subtle ways (happened in the past)

PR #74, opened 2022-08-24
Add tailwind css setup files and dependencies

### LINKED ISSUE #68: Station UI:  Setup a CSS Framework
Setting up a css framework as a baseline for station ui development with a specific theme would be a plus for future ui contributions. Following #46 discussion.
[Tailwind CSS](https://tailwindcss.com/docs/installation) would be a good choice.

jopedroliveira:
Based on #68, adding setup files for [Tailwind CSS](https://tailwindcss.com/docs/installation) as baseline for UI styling

Resolves #68


postcss.config.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
bajtos, line 1: Can we keep all frontend specific files inside `renderer` folder please? I.e. move `postcss.config.js` to `renderer/postcss.config.js`

Unless tooling like VSCode requires the config file to be in the project root.

What do you think, @jopedroliveira and @juliangruber?

The same comment applies to `tailwind.config.js` below.

jopedroliveira: Agree.

jopedroliveira: The docs refer that those files should be in the root. As I could check there's no clean way of moving it. 
I'll re-visit it later

bajtos: I see.

Quoting from https://vitejs.dev/guide/features.html#postcss
> If the project contains valid PostCSS config (any format supported by [postcss-load-config](https://github.com/postcss/postcss-load-config), e.g. postcss.config.js), it will be automatically applied to all imported CSS.

The docs for postcss-load-config offers many different file names for the configuration, but all of them are in the project root :(

Considering that we already have `vite.config.js` in project root too, I am fine to add PostCSS & Tailwind config files to project root too.

bajtos:
I am proposing to finish this PR as soon as possible, even if the design system is not complete. Focus on the project infrastructure in this PR, tweak the design later as part of #75.

/cc @jopedroliveira @juliangruber

jopedroliveira:
> I am proposing to finish this PR as soon as possible, even if the design system is not complete. Focus on the project infrastructure in this PR, tweak the design later as part of #75.
> 
> /cc @jopedroliveira @juliangruber

I was expecting to add general classes on this PR to setup the base for style. 
However, I agree that we can merge this to keep project infrastructure under control and iterate over css and stuff along feature development.

juliangruber (approved):
Merging this now to unblock @jopedroliveira. I'm not very experienced with tailwind, @bajtos a hindsight review would be cool :)

bajtos:
> I'm not very experienced with tailwind, @bajtos a hindsight review would be cool :)

I am not experienced with TailwindCSS either. I quickly skimmed through the changes and don't see any obvious problems.

It's great to have this PR merged 👍🏻

PR #102, opened 2022-09-06
chore: enable strict mode in JS files

bajtos:
- Configure eslint to require `use strict` in all JS files
- Add missing `use strict` lines where needed

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -19,7 +19,14 @@
   "overrides": [
     {
       "files": ["build/**/*.js", "main/**/*.js", "test/**/*.js", "*.js"],
+      "parserOptions": {
+        "sourceType": "commonjs"
+      },
+      "env": {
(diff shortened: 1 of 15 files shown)


.eslintrc.json
```diff
@@ -19,7 +19,14 @@
   "overrides": [
     {
       "files": ["build/**/*.js", "main/**/*.js", "test/**/*.js", "*.js"],
+      "parserOptions": {
+        "sourceType": "commonjs"
```
bajtos, line 23: I think auto-detection of source type does not work when using the TypeScript parser.

See https://github.com/silvermine/eslint-config-silvermine/issues/30#issuecomment-454897158 and https://github.com/eslint/typescript-eslint-parser/pull/583

juliangruber (approved):
`'use strict'`

PR #110, opened 2022-09-08
Activity Log backend

bajtos:
- feat: add ActivityLog backend
- feat: record ActivityLog entries

TODO:
- [x] IPC RPC API for getting initial activity entries
- [x] IPC Event for notifying about new activity
- [x] Limit the activity log to 100 entries


main/activity-log.js
```diff
@@ -0,0 +1,79 @@
+'use strict'
+
+const Store = require('electron-store')
+const configStore = new Store({
```
juliangruber, line 4:
```suggestion
const activityLogStore = new Store({
```

would this be a more fitting name?

main/activity-log.js
```diff
@@ -0,0 +1,79 @@
...
+      // We are storing ids as strings to allow us to switch to GUIDs in the future if needed
+      // When looking for the max id used, we need to cast strings to numbers.
+      .map(e => +e.id)
+      // Find the maximum id or return 0 when no events were recorded yet
+      .reduce((max, it) => it > max ? it : max, 0)
+  }
+
+  /**
+   * @param {import('./typings').ActivityEvent} args
+   * @returns {import('./typings').ActivityEntry}
```
juliangruber, line 24: would it be possible to import these types in the file header, or is this a limitation of our current type setup? I find these lines quite noisy

main/activity-log.js
```diff
@@ -0,0 +1,79 @@
...
+class ActivityLog {
+  #entries
+  #lastId
+
+  constructor () {
+    this.#entries = loadStoredEntries()
+    this.#lastId = this.#entries
+      // We are storing ids as strings to allow us to switch to GUIDs in the future if needed
+      // When looking for the max id used, we need to cast strings to numbers.
+      .map(e => +e.id)
```
juliangruber, line 17:
```suggestion
      .map(e => Number(e.id))
```

It's longer but I find this conveys the intention more clearly

main/activity-log.js
```diff
@@ -0,0 +1,79 @@
...
+      // Find the maximum id or return 0 when no events were recorded yet
+      .reduce((max, it) => it > max ? it : max, 0)
```
juliangruber, line 19:
```suggestion
    // We are storing ids as strings to allow us to switch to GUIDs in the future if needed
    // When looking for the max id used, we need to cast strings to numbers.
    // Find the maximum id or return 0 when no events were recorded yet
    this.#lastId = Math.max(...[0, this.#entries.map(e => Number(e.id))])
```

or

```suggestion
    this.#lastId = Math.max(...[
      // We are storing ids as strings to allow us to switch to GUIDs in the future if needed
      // When looking for the max id used, we need to cast strings to numbers.
      this.#entries.map(e => Number(e.id)),
      // Find the maximum id or return 0 when no events were recorded yet
      0
    ])
```

To me `Math.max()` is easier to reason about than `Array#reduce`. What do you think?

main/activity-log.js
```diff
@@ -0,0 +1,79 @@
...
+
+class ActivityLog {
+  #entries
+  #lastId
+
+  constructor () {
+    this.#entries = loadStoredEntries()
+    this.#lastId = this.#entries
+      // We are storing ids as strings to allow us to switch to GUIDs in the future if needed
+      // When looking for the max id used, we need to cast strings to numbers.
```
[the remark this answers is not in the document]
juliangruber: also, shouldn't the highest id always belong to the last entry, so no search is necessary?

main/activity-log.js
bajtos: I agree with you this is noisy.

You can import types, but it's a bit verbose, so I usually take the lazy approach you can see here. I will change my code to import those types so you can see which version you prefer.

BTW this DX issue has been discussed in  https://github.com/microsoft/TypeScript/issues/41825. Unfortunately, there was no solution yet.

main/activity-log.js
bajtos: A long time ago, in the Crankshaft days, `+val` was much more performant than `Number(val)`.  I am not sure if it's still the case.

Anyhow, this line is not on a critical path; I am okay with preferring readability over performance.

main/activity-log.js
bajtos: Agreed, I'll use the second suggestion 👍🏻

main/activity-log.js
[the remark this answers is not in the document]
bajtos: Yes, I am worried about collisions.

I guess you are right that we can simply use id of the last item as the last id 👍🏻 

> do we even needs ids? if the position is enough, we can address by array index

There is a race condition in the UI:
1. You want to subscribe for new activity entries  
2. You also want to fetch the list of existing entries. This is an asynchronous operation.

I think the correct implementation is this:
1. Subscribe to new activity
2. Get the existing activity

Now it's possible that an activity is recorded with such a timing that this entry triggers your event listener (1) and is also returned for your RPC call (2). The id allows the frontend to de-duplicate data coming from these two sources.

main/index.js
```diff
@@ -58,10 +71,23 @@ async function run () {
     await setupUpdater(ctx)
     await setupIpcMain()
 
+    ctx.recordActivity({ source: 'Station', type: 'info', message: 'Station started.' })
```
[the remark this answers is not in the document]
bajtos: I believe we report an error activity entry when the station cannot be started. With your suggestion, we will report "starting" followed by "cannot start". I feel it's just more noise.

I am sure this first version of Activity Log will need many tweaks until we reach the desired UX, both in Station and Saturn L2. Let's focus on laying down the fundamentals in this PR and then make any further improvements in follow-up pull requests.

WDYT?

main/saturn-node.js
```diff
@@ -41,7 +41,7 @@ async function setup (/** @type {import('./typings').Context} */ _ctx) {
     if (!childProcess) return
     stop()
   })
-  await start()
+  await start(ctx)
```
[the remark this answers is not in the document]
bajtos: Yeah, I am not sure how to feel about it either. We have inherited the backend architecture from ipfs-desktop. I don't have enough knowledge of Electron backends yet to design a better architecture confidently.

I am happy to revisit how we want to orchestrate backend components together. I prefer to do that a bit later if you don't mind. Maybe we can defer this until Lisbon, or perhaps hack on it together there?

main/activity-log.js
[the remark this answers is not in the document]
bajtos: See [8a759de](https://github.com/filecoin-project/filecoin-station/pull/110/commits/8a759de1fdff4643442bf8e258ba49d0914d2789)

main/activity-log.js
bajtos: See [8a759de](https://github.com/filecoin-project/filecoin-station/pull/110/commits/8a759de1fdff4643442bf8e258ba49d0914d2789)

main/typings.d.ts
```diff
@@ -1,4 +1,21 @@
...
+export interface ActivityEvent {
+  type: ActivityEventType;
+  source: ActivitySource;
+  message: string;
+}
+
+export interface ActivityEntry extends ActivityEvent {
+  id: string;
+  timestamp: number;
+}
```
bajtos, line 13: I am not entirely happy with these two names. They are very similar, I think it's not immediately clear what the difference is between `ActivityEvent` and `ActivityEntry`.

Can we find better names?

One option that comes to my mind is to rename `ActivityEntry` to something like `PersistedActivityEvent` or `StoredActivityEvent`.

WDYT, @juliangruber ?

renderer/src/main.tsx
```diff
@@ -14,3 +14,12 @@ ReactDOM.createRoot(
...
 )
+
+window.electron.getActivityLog().then(log => {
+  console.log('GOT INITIAL ACTIVITY LOG')
+  log.forEach(entry => console.log('[ACTIVITY] %j', entry))
+})
+
+window.electron.onActivityLogged(entry => {
+  console.log('[ACTIVITY] %j', entry)
+})
```
bajtos, line 25: This is a temporary UI to verify that the new infrastructure for ActivityLog works as intended.

Later, we can replace this either with some crude UI, or wait for the proper UI implementation from Subvisual.

I'd like to defer any UI work after this PR is landed.

main/activity-log.js
```diff
@@ -0,0 +1,73 @@
+'use strict'
+
+/** @typedef {import('./typings').ActivityEvent}  ActivityEvent */
+/** @typedef {import('./typings').ActivityEntry}  ActivityEntry */
```
bajtos, line 4: @juliangruber Would you like me to use this style of type-imports in all other places where I am referring to ActivityEvent and ActivityEntry types?

bajtos:
The PR is ready for final review.

bajtos:
@juliangruber Do you have any opinions on testing the renderer integration? 

Ideally, we should have end-to-end tests to verify that `window.electron.getActivityLog()` and `ndow.electron.onActivityLogged` work correctly. On the other hand, writing and maintaining e2e tests comes with extra overhead, I am not sure if it's a good ROI at this stage of our project.

main/activity-log.js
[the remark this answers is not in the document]
juliangruber: Makes sense!

> Yes, I am worried about collisions.

https://www.npmjs.com/package/monotonic-timestamp comes to mind for this, but I'm not sure if that actually improves our situation, as it tries to solve what we currently do with `.id` and `.timestamp` in one value.

> There is a race condition in the UI:

We might be able to fix this on the backend. Say the client only subscribes, never manually fetches the historical records. When the backend receives a subscription, it also pushes historical data over it. As long as the backend can keep items in sync, we wouldn't need IDs this way and could simply use array position.

main/index.js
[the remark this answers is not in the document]
juliangruber: I didn't see we had "cannot start" already, this is great as is! And +1 to improving the UX later on

juliangruber:
> @juliangruber Do you have any opinions on testing the renderer integration?
> 
> Ideally, we should have end-to-end tests to verify that `window.electron.getActivityLog()` and `ndow.electron.onActivityLogged` work correctly. On the other hand, writing and maintaining e2e tests comes with extra overhead, I am not sure if it's a good ROI at this stage of our project.

From my perspective we don't need tests for this at this stage. If anything we could start with IPC level tests, before looking at testing DOM results. And +1 to implementing both later.

juliangruber (changes requested):

main/activity-log.js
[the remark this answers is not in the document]
bajtos: > Say the client only subscribes, never manually fetches the historical records. When the backend receives a subscription, it also pushes historical data over it.

That's a cool design. 💯👍🏻 It will simplify frontend code a lot.

Do you know how to implement the part about detecting a subscription? We are using Electron IPC and Node.js EventEmitter APIs, I don't remember seeing a straightforward way for detecting new event subscriptions.

A solution that comes to my mind:
- Keep the current design where the frontend registers an event listener first and then sends an RPC call to the backend.
- By default, backend does not emit any events - it's in _paused_ state.
- Rework the RPC call from "get entries" to ask the backend to start emitting the activity entries instead. (Resume the stream.)

What do you think, @juliangruber? Do you have any other ideas on how to implement this part?

main/index.js
```diff
@@ -9,12 +9,26 @@ const setupUI = require('./ui')
...
+const { setupIpcMain, ipcMainEvents } = require('./ipc')
 const { setupAppMenu } = require('./app-menu')
 
+const { ActivityLog } = require('./activity-log')
+const { ipcMain } = require('electron/main')
+
 const inTest = (process.env.NODE_ENV === 'test')
 
+if (!app.isPackaged && !inTest) {
+  // Do not preserve old Activity entries in development mode
```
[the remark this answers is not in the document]
bajtos: +1 to use self-documenting variables

main/ui.js
```diff
@@ -66,8 +68,15 @@ module.exports = async function (ctx) {
   }
   ui.once('ready-to-show', ctx.showUI)
 
+  const onNewActivity = (/** @type {unknown[]} */ ...args) => {
+    if (ui.isDestroyed()) return // This happens when quitting the entire app
```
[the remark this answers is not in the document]
bajtos: Before this change, when I press `command + q`, the app crashed.

After taking another look, I think I understand the cause - we are unregistering "on window close" callbacks before the app quits. As a result, we don't unregister the `onNewActivity` listener and then try to send new activity (e.g. Saturn stopped) to a window that has already been destroyed.

https://github.com/filecoin-project/filecoin-station/blob/1723599bb6fe9bd832854839ae6129a353b06869/main/ui.js#L76-L82

main/activity-log.js
bajtos: Done in [4fd20c4](https://github.com/filecoin-project/filecoin-station/pull/110/commits/4fd20c419e7d3d584ae374e3c379aab4fce968fd).

I used typedefs only for the lines I am already changing to avoid mixing unrelated changes & extra code churn.

main/ui.js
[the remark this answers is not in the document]
bajtos: Fixed in [3271350](https://github.com/filecoin-project/filecoin-station/pull/110/commits/32713506cc3dec83fec1aea51c2d43223b0ade6e)

main/activity-log.js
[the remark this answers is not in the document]
bajtos: > do we even needs ids? if the position is enough, we can address by array index

When the Activity Log reaches the max size, we remove the oldest entry from the array. This changes the index of all entries from `X` to `X-1`. Each new event gets the same index `99`.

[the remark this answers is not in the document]
bajtos: Reworked activity streaming from `main` to `renderer` based on my description above - see 7e2dcad05

renderer/src/main.tsx
```diff
@@ -14,3 +14,11 @@ ReactDOM.createRoot(
...
   </React.StrictMode>
 )
+
+window.electron.onActivityLogged(entry => {
+  console.log('[ACTIVITY] %j', entry)
+})
+
+window.electron.resumeActivityStream().then(() => {
+  console.log('ACTIVITY STREAM RESUMED')
+})
```
bajtos, line 24: Please take a look at this example usage of the communication channel between the backend and the frontend for transmitting Activity Log entries. 

Is this new API (`onActivityLogged` and `resumeActivityStream`) easy enough to consume from a React frontend?

Are there any improvements we can make to make the integration with React simpler?

bajtos: Would it be helpful if I wrote down some documentation for this API, perhaps as JSDoc comments in `renderer/src/typings.d.ts` below?

main/activity-log.js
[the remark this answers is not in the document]
juliangruber: > What do you think, @juliangruber? Do you have any other ideas on how to implement this part?

I thought of it the same way, I think this will work well!

> When the Activity Log reaches the max size, we remove the oldest entry from the array. This changes the index of all entries from X to X-1. Each new event gets the same index 99.

With this new design, what do we even need IDs for? If we need them for the react `key={id}` prop, then we could assign random IDs. I don't see the need for monotonically increasing IDs with this design 🤔

renderer/src/main.tsx
juliangruber: - `onActivityLogged`: React hooks will want a way to unsubscribe as well, would it be simpler to let react register and unregister its own event listeners on the `ipc` object? Or if you still want to abstract it away, we could forward the events to a new event emitter, like this:


```js
window.electron.events.on('activity', () => {})
window.electron.events.off(...)
```

Could also call it `window.electron.activityLog`.

- `resumeActivityStream()`: This should work well as a `useEffect()`, however I'm also concerned about the clean up, and I think this method will need a matching `stopActivityStream()`.

What do you think about these function names, to make more clear what's happening?

```js
useEffect(() => {
  window.electron.activityLog.on('event', onEvent)
  window.electron.activityLog.subscribe()
  return () => {
    window.electron.activityLog.unsubscribe()
    window.electron.activityLog.removeEventListener('event', onEvent)
  }
})
```

Now that I read it, this pair of functions is awkward, shall we diverge from the EventEmitter pattern and implement it somewhat like this?

```js
useEffect(() => {
  const unsubscribe = window.electron.activityLog.subscribe(onEvent)
  return unsubscribe
})
```

juliangruber (changes requested):
see review comments above

main/activity-log.js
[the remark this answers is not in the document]
bajtos: > With this new design, what do we even need IDs for? If we need them for the react `key={id}` prop, then we could assign random IDs. I don't see the need for monotonically increasing IDs with this design 🤔

Yes, I was thinking about `key={id}` prop.

Does it actually matter how we are generating the id? What are your concerns about storing the `id` at backend?

renderer/src/main.tsx
bajtos: I am not sure if it's a good idea to subscribe & unsubscribe from inside `useEffect`. 

I know very little about React hooks, so I may be wrong, but my concern is that if React decides to re-run our component, and inside `useEffect` decides to create a new "hook instance", i.e. calls `unsubscribe` from the last instance and then creates a new instance by calling our `useEffect` callback, then we have a race condition. If a new activity is reported in the window between `unsubscribe` and `subscribe`, the we will not record it. Again, I am not sure if this is a possible scenario with React hooks.

IMO, we need the subscription to be long-lived, and keep the list of all Activity entries received so far in a persistent array (a global state).

I would think about this subscription as similar to a web socket connection. We don't want to subscribe/unsubscribe for web socket events from a component.

While searching for a solution, I found this article describing a solution based on `React.createContext`, which seems very reasonable to my beginner eyes.
https://aravindballa.com/writings/custom-hook-to-listen-websockets/

bajtos: I was thinking about implementing an `ActivityLogProvider` to store any long-living state & listeners, plus `useActivityLogSubscribe()` using `useContext` and `useEffect` for consumption by the actual app components.

```tsx
const App = () => (
  <ActivityLogProvider >
    {/* the actual app */}
    <ExampleComponent />
  </ActivityLogProvider >
);
```

bajtos: > If a new activity is reported in the window between `unsubscribe` and `subscribe`, the we will not record it. Again, I am not sure if this is a possible scenario with React hooks.

As I was thinking about this, I realized there won't be any race condition. Instead, when we subscribe after unsubscribing, the backend will re-emit all existing ActivityLog entries for us. So no data is lost, but it makes me wonder what will be the impact on performance.

bajtos: Setting my concerns aside, I really like your proposed API 👍🏻 

```ts
useEffect(() => {
  const unsubscribe = window.electron.activityLog.subscribe(onEvent)
  return unsubscribe
})
```

I would even go one step further (or maybe backward towards my current API):

```ts
useEffect(() => {
  const unsubscribe = window.electron.subscribeToActivityLog(onEvent)
  return unsubscribe
})
```

juliangruber: > So no data is lost, but it makes me wonder what will be the impact on performance.

If that data and ids are the same, React shouldn't re-render anything. And hooks are stable enough also not to re-fetch data unnecessarily.

Cool, looks like we're aligned!

main/activity-log.js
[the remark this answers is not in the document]
juliangruber: I'm absolutely for storing the id at the backend, but I think they can be random values and don't need to be monotonically increasing, since we don't need to use them for sorting any more, right? We can sort by timestamp, and iff two values have the same timestamp we could fall back to sorting these by their id. I think the system will be easier to reason about if ids are only used for identifying, and timestamps are used for sorting. Unless I miss something, in which case we could also have a quick call about this?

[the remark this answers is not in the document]
ronaldofs: is uuids an option for random ids? I think it's useful to have unique id for activity log entries.
for the UI, unique ids simplifies things (eg `key={id}`) 👍 
timestamps can help with ordering (although we can just read whatever order comes from backend).

renderer/src/main.tsx
ronaldofs: cool, we're using this pattern! (https://github.com/filecoin-project/filecoin-station/pull/123/files#diff-227719cb0a3e80022b8ead60744f738ee9689c17ce6edcc625d427e577b8460dR70)
and we can evolve it to a hook with react context  👍

main/activity-log.js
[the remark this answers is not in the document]
juliangruber: To help resolve this PR, I've made a PR demonstrating how I think this could work in a simpler way: https://github.com/filecoin-project/filecoin-station/pull/127

bajtos:
We did a pair programming session with @juliangruber today and brought the PR to the finish line.

@ronaldofs @jopedroliveira We implement a simple `ActivityLog` component that's consuming backend API and rendering a list of events. See `renderer/src/components/ActivityLog.tsx` and `renderer/src/App.tsx`. I hope this will make it super easy for you to integrate our changes. 

Feel free to rework both `App.tsx` and `ActivityLog.tsx` in any way you like. Our code is just a minimal working example to demonstrate that the backend integration is working.

main/activity-log.js
```diff
@@ -63,7 +63,7 @@ class ActivityLog {
  */
 function loadStoredEntries () {
   // A workaround to fix false TypeScript errors
-  return /** @type {any} */(activityLogStore.get('events', []))
+  return /** @type {any} */(activityLogStore.get('activities', []))
```
bajtos, line 66: This bug was tricky to find 🤓

bajtos: I think this was the reason why `ActivityLog.reset()` was not working in our pair-programming session, @juliangruber .

PR #127, opened 2022-09-14
prototype activity log layout

juliangruber:
To help clarify my reasoning in https://github.com/filecoin-project/filecoin-station/pull/110#discussion_r970526119, I decided to code it up.

@bajtos what do you think?

### DIFF
--- a/main/activity-log.js
+++ b/main/activity-log.js
@@ -1,31 +1,29 @@
 'use strict'
 
 /** @typedef {import('./typings').ActivityEvent}  ActivityEvent */
-/** @typedef {import('./typings').ActivityEntry}  ActivityEntry */
 
 const Store = require('electron-store')
+const crypto = require('node:crypto')
+
 const activityLogStore = new Store({
   name: 'activity-log'
 })
 
 class ActivityLog {
   #entries
-  #lastId
 
   constructor () {
     this.#entries = loadStoredEntries()
-    this.#lastId = Number(this.#entries.at(-1)?.id ?? 0)
   }
 
   /**
    * @param {ActivityEvent} args
-   * @returns {ActivityEntry}
+   * @returns {ActivityEvent}
    */
   recordEvent ({ source, type, message }) {
-    const nextId = ++this.#lastId
-    /** @type {ActivityEntry} */
+    /** @type {ActivityEvent} */
     const entry = {
-      id: String(nextId),
+      id: crypto.randomUUID(),
       timestamp: Date.now(),
       source,
       type,
@@ -61,7 +59,7 @@ class ActivityLog {
 }
 
 /**
- * @returns {ActivityEntry[]}
+ * @returns {ActivityEvent[]}
  */
 function loadStoredEntries () {
   // A workaround to fix false TypeScript errors
--- a/main/typings.d.ts
+++ b/main/typings.d.ts
@@ -2,14 +2,11 @@ export type ActivitySource = 'Station' | 'Saturn';
 export type ActivityEventType = 'info' | 'error';
 
 export interface ActivityEvent {
+  id: string;
+  timestamp: number;
   type: ActivityEventType;
   source: ActivitySource;
   message: string;
-}
-
-export interface ActivityEntry extends ActivityEvent {
-  id: string;
-  timestamp: number;
 }
 
 export interface Context {
--- a/renderer/src/main.tsx
+++ b/renderer/src/main.tsx
@@ -3,6 +3,7 @@ import ReactDOM from 'react-dom/client'
 import { BrowserRouter } from 'react-router-dom'
 import App from './App'
 import './index.css'
+import { ActivityEvent } from '../../main/typings'
 
 ReactDOM.createRoot(
   // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
@@ -15,7 +16,17 @@ ReactDOM.createRoot(
   </React.StrictMode>
 )
 
+const events: ActivityEvent[] = []
+
 window.electron.onActivityLogged(entry => {
+  events.push(entry)
+  if (events.length > 100) events.shift()
+
+  console.log('[EVENTS]', events.sort((a, b) => {
+    return a.timestamp !== b.timestamp
+      ? b.timestamp - a.timestamp
+      : a.id.localeCompare(b.id)
+  }))
   console.log('[ACTIVITY] %j', entry)
 })
 
--- a/renderer/src/typings.d.ts
+++ b/renderer/src/typings.d.ts
@@ -1,9 +1,9 @@
-import { ActivityEntry } from '../main/typings'
+import { ActivityEvent } from '../main/typings'
 
 export declare global {
   interface Window {
     electron: {
-      resumeActivityStream(): Promise<ActivityEntry[]>,
+      resumeActivityStream(): Promise<ActivityEvent[]>,
       onActivityLogged(callback: (activityEntry: ActivityEntry) => void),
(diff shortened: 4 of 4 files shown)


juliangruber:
The types are incomplete, will fix

main/activity-log.js
```diff
@@ -1,31 +1,30 @@
 'use strict'
 
 /** @typedef {import('./typings').ActivityEvent}  ActivityEvent */
-/** @typedef {import('./typings').ActivityEntry}  ActivityEntry */
+/** @typedef {import('./typings').RecordActivityOptions}  RecordActivityOptions */
```
juliangruber, line 4: I think we can simplify the types by thinking about them this way: One is the actual activity (`Activity`), the other is the options passed to `.recordActivity` (`RecordActivityOptions`). To me this is clearer than `event` and `entry`, and the variable names are nicely consistent.

juliangruber: We could also rename this to use `Event` instead of `Activity`, which is the more common term, but I do think here `Activity` works better, because it's clearer that we're referring to a product concept, and not a generic event.

main/typings.d.ts
```diff
@@ -1,19 +1,22 @@
...
   message: string;
 }
 
-export interface ActivityEntry extends ActivityEvent {
-  id: string;
-  timestamp: number;
+export interface RecordActivityOptions {
+  type: ActivityType;
+  source: ActivitySource;
+  message: string;
```
juliangruber, line 15: I removed the inheritance, because `Activity` is the main type, and `RecordActivityOptions` more a partial of it.

renderer/src/main.tsx
```diff
@@ -15,8 +16,18 @@ ReactDOM.createRoot(
...
+
+window.electron.onActivityLogged(activity => {
+  activities.push(activity)
+  if (activities.length > 100) activities.shift()
+
+  console.log('[ACTIVITIES]', activities.sort((a, b) => {
+    return a.timestamp !== b.timestamp
+      ? b.timestamp - a.timestamp
+      : a.id.localeCompare(b.id)
+  }))
```
juliangruber, line 29: we could also move the sorting logic to the main thread, I don't think it makes a difference

renderer/src/main.tsx
```diff
@@ -15,8 +16,19 @@ ReactDOM.createRoot(
...
-  console.log('[ACTIVITY] %j', entry)
+const activities: Activity[] = []
+
+window.electron.onActivityLogged(activity => {
+  activities.push(activity)
+  activities.sort((a, b) => {
+    return a.timestamp !== b.timestamp
+      ? b.timestamp - a.timestamp
+      : a.id.localeCompare(b.id)
+  })
```
juliangruber, line 27: if we moved sorting to the backend, then we wouldn't have to sort on every new event. Otherwise I think we could do it in either location, and having it on the view layer might be easier to get right wrt race conditions

juliangruber:
> Error: main/test/activity-log.test.js(55,5): error TS2683: 'this' implicitly has type 'any' because it does not have a type annotation.
Error: Process completed with exit code 2.

@bajtos I don't understand why this isn't showing up in main ci 🤔

main/activity-log.js
bajtos: I like the idea to user `Activity` to describe the data structure we are storing in our log 👍🏻 

I have mixed feelings about the name `RecordActivityOptions`. In my mind, options is for optional configuration tweaking the behaviour of a function. How about something like `RecordActivityFields` or `NewActivityProps`?

main/activity-log.js
```diff
@@ -1,47 +1,46 @@
...
-   * @returns {ActivityEntry}
+   * @param {RecordActivityOptions} args
+   * @returns {Activity}
    */
-  recordEvent ({ source, type, message }) {
-    const nextId = ++this.#lastId
-    /** @type {ActivityEntry} */
-    const entry = {
-      id: String(nextId),
+  recordActivity ({ source, type, message }) {
```
bajtos, line 24: Since this is a method of `ActivityLog` class, maybe we can shorten the name to simply `record`?

```suggestion
  record ({ source, type, message }) {
```

main/typings.d.ts
bajtos: It makes sense to remove inheritance 👍🏻 

An alternative approach that does not duplicate property definitions:

```ts
export type RecordActivityOptions = Pick<Activity, 'type' | 'source' | 'message'>;
```

renderer/src/main.tsx
```diff
@@ -15,8 +16,19 @@ ReactDOM.createRoot(
...
-window.electron.onActivityLogged(entry => {
-  console.log('[ACTIVITY] %j', entry)
+const activities: Activity[] = []
+
+window.electron.onActivityLogged(activity => {
+  activities.push(activity)
+  activities.sort((a, b) => {
+    return a.timestamp !== b.timestamp
+      ? b.timestamp - a.timestamp
+      : a.id.localeCompare(b.id)
```
bajtos, line 26: Are you sure ids created by `crypto.randomUUID()` monotonically increasing over time?

Maybe the order of events reported in the same millisecond does not matter and that's why we can fallback our sorting comparisons to comparing ids?

I think it would be great to add a code comment to clarify this.

main/activity-log.js
juliangruber: Sounds good, I tried to look up prior art on this and found that the types for node for example call arguments `options` but their type `...Args`:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3b2cb0f873d4caca3e4fc01aeaf9f8ec40852d8f/types/node/http.d.ts#L138-L139

Either would work for me 👍

main/typings.d.ts
juliangruber: cool I like that!

main/activity-log.js
juliangruber: good idea!

renderer/src/main.tsx
juliangruber: They don't increase, and that's not an issue in my mind. As long as we sort deterministically, no human will have noticed that the events from the same millisecond have been mixed up. And we sort deterministically by giving events IDs when storing them. I'll add a comment 👍

juliangruber: alternative idea: Is there a timing api with such precision that it's guaranteed that two successive calls won't get the same values?

juliangruber: `process.hrtime()` _seems_ to work:

```
> [process.hrtime(), process.hrtime()]
[ [ 845635, 876114697 ], [ 845635, 876117208 ] ]
```

main/typings.d.ts
bajtos: On the second thought, this is probably more future proof when it comes to adding new Activity props:

```ts
export type RecordActivityOptions = Omit<Activity, 'id' | 'timestamp'>;

renderer/src/main.tsx
bajtos: > As long as we sort deterministically, no human will have noticed that the events from the same millisecond have been mixed up. 

Unless these two events happen to be something like "Starting the station" and "Starting Saturn module". It would look weird to have them in the reverse direction.

Anyhow, I think this does not really matter, especially not now.

bajtos:
Let's merge this into my branch and then work together there.

main/activity-log.js
bajtos: Let's use `NewActivityArgs` then.

Issue #81, opened 2022-08-30
Expose stats & events from L2-node to WebUI


bajtos:
Station backend needs to expose the following stats & events for the WebUI:

## Stats

Stats are persisted by L2 Node, we just need to call HTTP API to get the new numbers.

- [x] Number of jobs served
  - We should periodically call L2-node's HTTP endpoint `/stats` and parse the number from there, see https://github.com/filecoin-saturn/L2-node/pull/59

👉🏻 Done in https://github.com/filecoin-project/filecoin-station/pull/136

## Activity feed 

Activity is reported via events, we need to persist them on our side.
- [x] Implement persistence for activity feed. Keep X latest entries only and delete old ones.

Record the following events
- [x] Saturn L2 Node started
- [x] Saturn L2 Node exited or panicked (include the panic message?)
- [x] Any lines from L2-node's stdout stream that are prefixed with `INFO`

👉🏻 Done in https://github.com/filecoin-project/filecoin-station/pull/110

## IPC integration

Ideally, we should have two APIs for WebUI:

1. Get the initial state (# of jobs served, activity log)
2. Subscribe to events to get notified when a new activity is reported or more jobs were finished

juliangruber:
> [ ] Implement persistence for activity feed. Keep X latest entries only and delete old ones.

Do you have an idea how to implement this? For now we could just keep state as an array and flush to disk as JSON when there is an update, debounced every second?

juliangruber:
I think that works!

PR #136, opened 2022-09-20
feat: show total jobs processed

bajtos:
See https://github.com/filecoin-project/filecoin-station/issues/81

station-jobs

- Modify Saturn integration to parse `API` URL from the stdout instead of `WebUI`
- Periodically call Saturn's `stats` API to obtain the number of successful retrievals. The interval is 100ms.
- Implement a `JobCounter` helper to record per-module jobs, persist them across restarts, and calculate the total number of jobs handled.
- Add IPC integration allowing the renderer to obtain the number of jobs handled and subscribe to updates. 
- Add a simple React component with a minimal UI to visibly demonstrate the new feature.

~~The backend emits the "job counter updated" event every 30ms. I decided to use a dedicated interval to better support the future when we are periodically fetching job counters from multiple modules, in which case emitting "job counter updated" for each module could be too often. (E.g. we get stats from module 1 in time X, then module 2 returns with stats in time X+2ms. If we emit the event on every change, then we would emit two events within 2ms.)~~

~~I don't have a strong opinion on whether this is actually a better implementation and I am happy to consider different options. For example, we can follow the ActivityLog design where a new activity emits an IPC event and the UI component forward these events to the renderer window if the UI is shown.~~


main/index.js
```diff
@@ -14,6 +14,7 @@ const { setupAppMenu } = require('./app-menu')
 
 const { ActivityLog } = require('./activity-log')
 const { ipcMain } = require('electron/main')
+const { JobCounter } = require('./job-counter')
```
bajtos, line 17: I did not spend too much time picking the name "job counter". I am happy to use a more fitting name if we can find one.

main/typings.d.ts
```diff
@@ -11,9 +11,14 @@ export interface Activity {
...
 
+export type ModuleJobCount = Record<string, number>;
+
 export interface Context {
   recordActivity(activity: RecordActivityArgs): void;
-  getAllActivities(): void;
+  getAllActivities(): Activity[];
+
+  recordModuleJobCount(moduleName: string, count: number): void;
+  getNumberOfAllJobsProcessed(): number;
```
bajtos, line 21: Similarly here. I am happy to use different names if we can better ones.

jopedroliveira (approved):
I would like this to follow the same approach as activity log but that's not an issue at all. We can move forward with this approach and review it later.

bajtos:
> I would like this to follow the same approach as activity log but that's not an issue at all. We can move forward with this approach and review it later.

Thanks for chiming in, @jopedroliveira!

From the renderer (frontend) perspective, there is no difference between ActivityLog and JobCounter. You have two APIs: one for getting the initial value, another for subscribing to updates.

The difference lays in backend and how it decides when to emit "update" events.
- ActivityLog emits the event whenever a new activity is recorded
- JobCounter emits the event at a regular interval.

I am going to try to rework JobCounter to use the same design as ActivityLog, I'll push the changes in a standalone commit.

main/index.js
```diff
@@ -63,6 +66,12 @@ const ctx = {
     ipcMain.emit(ipcMainEvents.ACTIVITY_LOGGED, activityLog.getAllEntries())
   },
 
+  getNumberOfAllJobsProcessed: () => jobCounter.getNumberOfAllJobsProcessed(),
```
bajtos, line 69: Maybe `getJobCount` or `getTotalJobCount` would be a better name?

main/ipc.js
```diff
@@ -9,6 +9,7 @@ const stationConfig = require('./station-config')
 
 const ipcMainEvents = Object.freeze({
   ACTIVITY_LOGGED: 'station:activity-logged',
+  JOB_COUNTER_UPDATED: 'station:job-count-updated',
```
juliangruber, line 12: As mentioned above, the job-counter vs job-count is a bit confusing

main/job-counter.js
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * @returns {import('./typings').ModuleJobCount}
+ */
+function loadStoredCounters () {
+  // A workaround to fix false TypeScript errors
+  return /** @type {any} */(jobCounterStore.get('counters', {}))
+}
+
+/**
+ * @param {import('./typings').ModuleJobCount} counters
```
juliangruber, line 48: could we move the type imports to the file header, for readability?

main/job-counter.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+function storeCounters (counters) {
+  jobCounterStore.set('counters', counters)
+}
+
+module.exports = {
+  JobCounter
```
juliangruber, line 55: why not make this the default export?

renderer/src/App.tsx
```diff
@@ -30,6 +31,7 @@ function Home (): JSX.Element {
   return (
     <div style={{ marginTop: '2em' }}>
       <h2>Welcome to Filecoin Station</h2>
+      <JobCounter />
```
juliangruber, line 34: To match the visible UI, shall we rename this component either to `TotalJobsCompleted` or `JobStats`?

main/index.js
juliangruber, line 17: What do you think about calling this `JobStats` instead? This way we can add additional stats in the future (like job execution time or error rate) and don't have the awkwardly similar but different events `JOB_COUNTER_UPDATED` and `job-count-updated`.

juliangruber: Also it makes more sense to me to call this stat something like `totalJobsCompleted` or `jobsCompleted`, because we're counting the completed jobs and not the overall jobs (technically). This way UI and variables are aligned.

main/index.js
```diff
@@ -63,6 +66,12 @@ const ctx = {
     ipcMain.emit(ipcMainEvents.ACTIVITY_LOGGED, activityLog.getAllEntries())
   },
 
+  getTotalJobCount: () => jobCounter.getTotalJobCount(),
```
juliangruber, line 69: If you agree with the above comment, `getTotalJobsCompleted` would fit here nicely for example

main/saturn-node.js
```diff
@@ -106,16 +113,21 @@ async function start (/** @type {Context} */ ctx) {
...
-      if (webuiMatch) {
-        webUrl = webuiMatch[1]
+      const apiMatch = output.match(/^API: (http.*)$/m)
+      if (apiMatch) {
+        apiUrl = apiMatch[1]
 
         appendToChildLog('Saturn node is up and ready')
-        console.log('Saturn node is up and ready (Web URL: %s)', webUrl)
+        console.log('Saturn node is up and ready (API URL: %s)', apiUrl)
+        webUrl = apiUrl + 'webui'
```
juliangruber, line 122:
```suggestion
        webUrl = `${apiUrl}webui`
```

modernize all the things

main/saturn-node.js
```diff
@@ -4,6 +4,7 @@ const path = require('path')
 const fs = require('node:fs/promises')
 const { setTimeout } = require('timers/promises')
 const assert = require('node:assert')
+const { request } = require('undici')
```
juliangruber, line 7: could we simplify the backend by also using fetch, the JS standard api?

main/saturn-node.js
```diff
@@ -239,6 +253,17 @@ function handleActivityLogs (ctx, text) {
     })
 }
 
+async function fetchJobCount (/** @type {Context} */ ctx) {
+  const res = await request(apiUrl + 'stats', { throwOnError: true })
+  const stats = await res.body.json()
+  const jobCount = stats.NSuccessfulRetrievals
```
juliangruber, line 259:
```suggestion
  const jobCount = stats?.NSuccessfulRetrievals
```

Just in case, since we're also defensive below

main/ui.js
```diff
@@ -86,10 +83,30 @@ module.exports = async function (ctx) {
   // that were added to keep app running when the UI is closed.
   // (Without this, the app would run forever and/or fail to update)
   app.on('before-quit', () => {
-    ipcMain.removeListener(ipcMainEvents.ACTIVITY_LOGGED, onNewActivity)
+    stopIpcEventForwarding()
```
juliangruber, line 86: is this actually necessary at all?

juliangruber (changes requested):
I meant to hit "Request changes"

main/job-counter.js
bajtos: Of course.

I am usually adding these type names via VS Code IntelliSense, that's why we are getting these inline imports. 🤷🏻

main/job-counter.js
bajtos: I prefer to use named exports as they are more future-proof. Adding the second named export requires no changes to existing code. Switching from a single default export to two named export requires changes in all places that were using the default export.

main/saturn-node.js
bajtos: We can use fetch if we are ok with the ramifications:
- Fetch is not a built-in API in Node.js 16. We will keep using Undici, but with a different API (`fetch` vs `request`).
- The performance of the `fetch` API in both Undici and Node.js 18 is much worse than Undici's `request` API due to WebStreams. Our current request/fetch API usage is not performance critical; thus, it's ok to prefer DX over performance.

main/job-counter.js
juliangruber: ok!

main/saturn-node.js
juliangruber: +1 to using fetch via undici

Do you know how much slower fetch is?

main/ui.js
bajtos: Yes. When we don't unsubscribe our IPC event listeners, they get triggered while the app is shutting down. Our handler forwards the event to `ui`, which has already been destroyed. Unhandled rejection is triggered, the app shutdown does not complete, and the app keeps running.

main/index.js
bajtos: > What do you think about calling this JobStats instead?

I love that idea! 💯

bajtos: > call this stat something like `totalJobsCompleted`

🎱

main/saturn-node.js
bajtos: I am not convinced this is an improvement - how about micro-performance? - but I am okay with following your style.

main/saturn-node.js
bajtos: I am not sure. If I am reading this comment from Feb 2022 correctly, then `fetch` was 8x slower than `request`.
https://github.com/nodejs/undici/issues/1203#issuecomment-1033189710

The Undici team has landed several performance improvements since then, but the real problem is in WebStreams. Quoting from https://github.com/nodejs/undici/issues/1203#issuecomment-1100969210:
> However, the major bottleneck is under WebStreams implementation in Node.js core, thus, I don't think there is a clear path for undici.fetch to improve its performance significantly without touching Node.js internals.

main/ui.js
bajtos: Done in https://github.com/filecoin-project/filecoin-station/pull/136/commits/66cab6f94a376971d952e936a010fa34a3e8b309.

I find the `request` API slightly more ergonomic than the `fetch` API, but I guess there is more value in using a standard Web API that more people are familiar with.

main/saturn-node.js
bajtos: Done in https://github.com/filecoin-project/filecoin-station/pull/136/commits/66cab6f94a376971d952e936a010fa34a3e8b309.

I find the `request` API slightly more ergonomic than the `fetch` API, but I guess there is more value in using a standard Web API that more people are familiar with.

main/index.js
```diff
@@ -63,6 +66,12 @@ const ctx = {
     ipcMain.emit(ipcMainEvents.ACTIVITY_LOGGED, activityLog.getAllEntries())
   },
 
+  getTotalJobsCompleted: () => jobStats.getTotalJobsCompleted(),
+  setModuleJobsCompleted: (moduleName, count) => {
+    jobStats.setModuleJobsCompleted(moduleName, count)
+    ipcMain.emit(ipcMainEvents.JOB_STATS_UPDATED, jobStats.getTotalJobsCompleted())
```
juliangruber, line 72:
```suggestion
    ipcMain.emit(ipcMainEvents.JOB_STATS_UPDATED, {
      totalJobsCompleted: jobStats.getTotalJobsCompleted()
    })
```

I think this would be nice to make way for further stats, also this way the content matches the event name more

main/saturn-node.js
```diff
@@ -239,6 +253,28 @@ function handleActivityLogs (ctx, text) {
...
+
+  /** @type {any} */
+  const stats = await res.json()
+
+  const jobCount = stats?.NSuccessfulRetrievals
+  if (typeof jobCount !== 'number') {
+    const msg = 'Unexpected stats response - NSuccessfulRetrievals is not a number. Stats: ' + JSON.stringify(stats)
+    throw new Error(msg)
+  }
+  ctx.setModuleJobsCompleted('saturn', jobCount)
```
juliangruber, line 271:
```suggestion
  const jobsCompleted = stats?.NSuccessfulRetrievals
  if (typeof jobsCompleted !== 'number') {
    const msg = 'Unexpected stats response - NSuccessfulRetrievals is not a number. Stats: ' + JSON.stringify(stats)
    throw new Error(msg)
  }
  ctx.setModuleJobsCompleted('saturn', jobsCompleted)
```

renderer/src/components/TotalJobsCompleted.tsx
```diff
@@ -0,0 +1,31 @@
...
+  // TODO: add proper CSS styling :)
+  const style: React.CSSProperties = {
+    marginLeft: '2em',
+    fontSize: '2em',
+    fontFamily: 'monospace'
+  }
+
+  return <>
+    <h3>Total jobs completed</h3>
+    <p style={style}>{totalJobCount}</p>
```
juliangruber, line 29:
```suggestion
  const [totalJobsCompleted setTotalJobsCompleted] = useState<number>(0)

  useEffect(() => {
    (async () => {
      const jobsCompleted = await window.electron.getTotalJobsCompleted()
      setTotalJobsCompleted(jobsCompleted)
    })()
  }, [])

  useEffect(() => {
    const unsubscribe = window.electron.onJobStatsUpdated(count => {
      setTotalJobsCompleted(count)
    })
    return unsubscribe
  }, [])

  // TODO: add proper CSS styling :)
  const style: React.CSSProperties = {
    marginLeft: '2em',
    fontSize: '2em',
    fontFamily: 'monospace'
  }

  return <>
    <h3>Total jobs completed</h3>
    <p style={style}>{totalJobsCompleted}</p>
```

I find consistent naming really helpful

main/ui.js
juliangruber: > Unhandled rejection is triggered, the app shutdown does not complete, and the app keeps running.

Ok! It's better to not have an unhandled rejection then

juliangruber (changes requested):
🏎️

main/index.js
bajtos: I have considered emitting stats as an object too. But then, we would also need to change the getter function for consistency.

In the end, I decided to keep things simple. We have complete control of the backend & frontend, and they are both moving together in a lock step. Reworking stats from a number to an object can be easily done at any time.

[YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

renderer/src/components/TotalJobsCompleted.tsx
bajtos: Thank you! I missed this part; I agree it's much better renamed.

juliangruber (approved):
SHIP IT 😁

Issue #78, opened 2022-08-30
UI error monitoring


bajtos:
Setup error monitoring for Station UI.

- [x] Setup a new Sentry project (possibly using an existing PL account)
  → https://sentry.io/protocol-labs-ip/filecoin-station/ (not public)
- [x] Integrate Sentry into Station frontend
  → https://github.com/filecoin-project/filecoin-station/pull/138
- [x] Add a new paragraph to the Consent screen to describe what data is sent to Sentry. If the consent screen is not implemented yet, then add a new subtask to the appropriate issue.
  → https://github.com/filecoin-project/filecoin-station/pull/140
- [x] Report new releases to Sentry - see https://docs.sentry.io/product/releases/setup/release-automation/github-actions/
  → https://github.com/filecoin-project/filecoin-station/pull/163
- [x] Upload source maps to Sentry - see https://docs.sentry.io/platforms/javascript/sourcemaps/
  → https://github.com/filecoin-project/filecoin-station/pull/163
- [x] Disable Sentry integration for local development
  → https://github.com/filecoin-project/filecoin-station/pull/183
- [x] (should have) use `production` environment for builds created by GHA for git tags, `development` in all other cases.
  → done by the combination of #163 and #183 and #212
- [x] Consider integrating Sentry into Station backend too
  → https://github.com/filecoin-project/filecoin-station/pull/184

juliangruber:
ping @patrickwoodhead

patrickwoodhead:
@bajtos - I have create a Sentry team and sent you and @juliangruber an invitation

bajtos:
> use `production` environment for builds created by GHA for git tags, development in all other cases

Here is my understanding of our current setup:

1. Inside `package.json`, both `buildTag` and `buildNumber` are set to null. As a result, CI tests running unpackaged code consider the version to be a dev one and disable Sentry integration. Developers running `npm start` and `npm test` locally fall into this category too.

2. When people download installers produced by our CI builds:

   - CI builds running for a git tag update `package.json` fields in the produced packages to report SENTRY RELEASE `x.y.z`. 
  
     One of our build steps configures Sentry with `environment=production` for this release.

   - Other CI builds triggered from our repository update `package.json` fields in the produced packages to report SENTRY RELEASE `x.y.z.{build-number}`. 
   
     One of our build steps configures Sentry with `environment=development` for this release. 

[...]

PR #138, opened 2022-09-20
feat: basic Sentry error monitoring

bajtos:
See #78

- feat: setup Sentry error monitoring
- feat: add release to Sentry config

Out of scope:
- Report releases and source maps to Sentry
- Report errors from the backend

### DIFF
--- a/main/index.js
+++ b/main/index.js
@@ -21,11 +21,8 @@ const setupUpdater = require('./updater')
 const inTest = (process.env.NODE_ENV === 'test')
 const isDev = !app.isPackaged && !inTest
 
-console.log('BUILD VERSION', BUILD_VERSION)
-
-// Expose additional metadata for Electron preload script
-process.env.IS_PACKAGED = app.isPackaged ? '1' : undefined
-process.env.STATION_VERSION = BUILD_VERSION
+console.log('Filecoin Station build version:', BUILD_VERSION)
+process.env.STATION_BUILD_VERSION = BUILD_VERSION
 
 function handleError (/** @type {any} */ err) {
   ctx.recordActivity({
--- a/main/index.js
+++ b/main/index.js
@@ -3,8 +3,8 @@
 const { app, dialog } = require('electron')
 const { ipcMainEvents, setupIpcMain } = require('./ipc')
 const { ActivityLog } = require('./activity-log')
+const { BUILD_VERSION } = require('./consts')
 const { JobStats } = require('./job-stats')
-const { STATION_VERSION } = require('./consts')
 const { ipcMain } = require('electron/main')
 const log = require('electron-log')
 const path = require('node:path')
@@ -21,12 +21,11 @@ const setupUpdater = require('./updater')
 const inTest = (process.env.NODE_ENV === 'test')
 const isDev = !app.isPackaged && !inTest
 
+console.log('BUILD VERSION', BUILD_VERSION)
+
 // Expose additional metadata for Electron preload script
 process.env.IS_PACKAGED = app.isPackaged ? '1' : undefined
-process.env.STATION_VERSION = app.isPackaged
-  ? STATION_VERSION
-  // TODO: can we use Git commit SHA instead?
-  : `dev-on-${STATION_VERSION}`
+process.env.STATION_VERSION = BUILD_VERSION
 
 function handleError (/** @type {any} */ err) {
   ctx.recordActivity({
--- a/main/index.js
+++ b/main/index.js
@@ -22,6 +22,8 @@ const inTest = (process.env.NODE_ENV === 'test')
 const isDev = !app.isPackaged && !inTest
 
 console.log('Filecoin Station build version:', BUILD_VERSION)
+
+// Expose additional metadata for Electron preload script
 process.env.STATION_BUILD_VERSION = BUILD_VERSION
 
 function handleError (/** @type {any} */ err) {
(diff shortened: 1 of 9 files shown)


main/index.js
```diff
@@ -21,6 +22,13 @@ const { ipcMain } = require('electron/main')
 const inTest = (process.env.NODE_ENV === 'test')
 const isDev = !app.isPackaged && !inTest
 
+// Expose additional metadata for Electron preload script
```
juliangruber, line 25: Is overwriting `process.env` the best way to pass data to the client? 🤔

main/index.js
```diff
@@ -21,6 +22,13 @@ const { ipcMain } = require('electron/main')
...
+process.env.IS_PACKAGED = app.isPackaged ? '1' : undefined
+process.env.STATION_VERSION = app.isPackaged
+  ? STATION_VERSION
+  // TODO: can we use Git commit SHA instead?
```
juliangruber, line 29: We probably could, but is that a meaningful change for now?

renderer/src/main.tsx
```diff
@@ -1,9 +1,24 @@
...
+  dsn: 'https://ff9615d8516545158e186d863a06a0f1@o1408530.ingest.sentry.io/6762462',
+  integrations: [new BrowserTracing()],
+
+  release: window.electron.stationMetadata.version,
+  environment: window.electron.stationMetadata.environment,
+
+  // Set tracesSampleRate to 1.0 to capture 100%
+  // of transactions for performance monitoring.
+  // We recommend adjusting this value in production
+  tracesSampleRate: 1.0
```
juliangruber, line 19: submitting traces is expensive (and potentially intrusive), because a lot of data will be captured and submitted. Shall we reduce this right away, to say 10%?

main/index.js
juliangruber: Tbh I haven't thought about how to do this in a setup where the renderer doesn't have access to `require('electron')`

renderer/src/main.tsx
bajtos: Good point. Do we actually need tracing at all? Maybe we can start with collecting errors only and wait with adding tracing until we need it?

main/index.js
bajtos: It depends on what we want to see in the Sentry dashboard when looking at errors. At the moment, when running the Station via `npm start`, errors are labelled with release `dev-on-0.5.0`. If we think that's good enough, then I can remove this TODO comment.

I consider this as a longer-term idea, not something to work on right now. I think we would need to obtain the Git commit SHA in a sync way (during the first tick of the event loop), which is tricky.

main/index.js
bajtos: The preload script is intentionally sandboxed for security reasons. It has access to a very limited set of APIs, see https://www.electronjs.org/docs/latest/tutorial/tutorial-preload

I am open to use a different mechanism for passing data from `main` to `preload`, what do you suggest?

juliangruber: I don't have anything on the top of my mind, I was more curious if there were well known alternatives. Thanks for the context!

renderer/src/main.tsx
juliangruber: I'm concerned that we'll never start using traces if we don't at least have _some_ in our dashboard 🤔 But I don't feel strongly about this either, we could just remove them too

main/index.js
juliangruber: Since we can annotate dev environments in Sentry I don't think we have to remove Sentry from dev - it won't pollute our production date - but at the same time removing it might be the simpler path here?

bajtos: Great discussion!

I see three different environments to cater for:
- Production - packaged builds produced from git tags (releases).
- Testing - automated CI tests run via GitHub Actions. I think we run the app unpackaged.
- Local development

I agree we don't need Sentry for the dev env because we can inspect errors in DevTools.

I think it's potentially useful to collect error reports from our CI runs. Ideally, these errors should be reported by our test suite. Considering the effort required to do that, I feel it's much easier to use Sentry.

renderer/src/main.tsx
bajtos: Makes sense. I'll set the rate to `0.1` then. We can easily tweak this later, based on what we learn from real-world usage.

main/index.js
juliangruber: Uh that's a great idea! CI can be such a PITA to debug

bajtos:
I reworked the Sentry configuration as follows:

- Changed `tracesSampleRate` to `0.1`

- Removed `environment` config. I'd like to configure this in such way that only release build produced for tags are reporting with `production` environment and everything else reports with `development`. It's not straightforward, let's leave it out of the scope of this PR.

- Reworked the version reported to Sentry as follows:
   - Release builds report `x.y.z`, e.g. `0.5.0`
   - Non-release builds from GitHub Actions report `x.y.z.{build-number}`, e.g. `0.5.0.444`
   - Local/dev builds report `x.y.z.1-dev`, e.g. `0.5.0.1-dev`

@juliangruber please review again

main/index.js
bajtos: The new version I pushed today, we are reporting all errors to Sentry. Errors from local dev builds can be distinguished via version number ending with `.1-dev`.

Do you agree this is good enough for now, or should I look into ways how to disable error reporting from dev builds?

juliangruber (approved):
This is great! Thank you for putting the extra thought into this!

PR #139, opened 2022-09-20
eslint: add `"sort-imports"`

juliangruber:
https://eslint.org/docs/latest/rules/sort-imports


bajtos (commented):
Nice! 👏🏻 

I am surprised we didn't need to fix any JavaScript files, only TypeScript. 

Could you please verify that ESLint is applying this rule for CommonJS imports too?

Maybe we need to add something like this? https://github.com/zcuric/eslint-plugin-require-sort

Should we enable `sort-import` rule at the top level, when it applies only to `renderer` files? Maybe it's better to move that rule to `"files": ["renderer/**/*"]` section at the bottom? I can see arguments for both ways.

juliangruber:
Great catch, and great suggestion. Implemented 🙏

bajtos (approved):
👏🏻

Issue #70, opened 2022-08-24
Station UI: Onboarding and Login page


jopedroliveira:
- [x] Onboarding - Intro screens
- [x] Ask for consent or quit app
  - [x] Data collection - see #118 
  - [x] Using their computer & network resources      
- [x] Connect FIL Wallet

PR #140, opened 2022-09-20
feat: user onboarding and consent

jopedroliveira:
This PR implements the Station onboarding screen with modal steps ending on the user consent ask.

See #70


juliangruber (changes requested):

renderer/src/components/InterfaceCalls.tsx
```diff
@@ -0,0 +1,7 @@
+export async function getStationUserOnboardingCompleted (): Promise<boolean> {
+  return await window.electron.stationConfig.getOnboardingCompleted()
+}
+
+export async function setStationUserOnboardingCompleted (): Promise<void> {
+  return await window.electron.stationConfig.setOnboardingCompleted()
+}
```
juliangruber, line 7: I'm not sure about this file, and its placement. It's not a frontend component, is it?

What do you think about moving it to `renderer/src/lib/station-config.tsx`, to mirror the file `main/station-config.js`? And it would be cool to mirror its function names as well (`getOnboardingCompleted` and `setOnboardingCompleted`), to unify the code base.

renderer/src/components/Onboarding.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { ReactComponent as CurrentPage } from './../assets/img/paginator-current.svg'
+
+interface IFooter {
+  page: number,
+  pages: number,
+  next: () => void,
+  prev: () => void
+}
+
+const Footer: FC<IFooter> = ({ page, pages, next, prev }) => {
```
juliangruber, line 13: what do you think about naming interfaces like this `FooterProps`, to match what we already have in for example `FilAddressForm.tsx`?

```suggestion
interface FooterProps {
  page: number,
  pages: number,
  next: () => void,
  prev: () => void
}

const Footer: FC<FooterProps> = ({ page, pages, next, prev }) => {
```

I think in general this is a more common naming

renderer/src/components/Onboarding.tsx
```diff
@@ -0,0 +1,108 @@
...
+  return (
+    <div className="flex justify-between">
+      <button
+        className="btn-secondary-small "
```
juliangruber, line 17:
```suggestion
        className="btn-secondary-small"
```

renderer/src/components/Onboarding.tsx
```diff
@@ -0,0 +1,108 @@
...
+        disabled={page === 0}
+        onClick={prev}>
```
juliangruber, line 19:
```suggestion
        onClick={prev}
        type="button">
```

For accessibility

renderer/src/components/Onboarding.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <span className="">Back</span>
```
juliangruber, line 20:
```suggestion
        <span>Back</span>
```

renderer/src/components/Onboarding.tsx
```diff
@@ -0,0 +1,108 @@
...
+      </button>
+      <div className='flex flex-row items-center space-between gap-3'>
+        {[...Array(pages - 1).keys()].map((_, index) => {
+          return index === page ? <CurrentPage key={index}/> : <Page key={index}/>
+        })}
+      </div>
+      <button
+        className="btn-secondary-small "
+        onClick={next}>
```
juliangruber, line 29:
```suggestion
        onClick={next}
        type="submit">
```

renderer/src/components/Onboarding.tsx
```diff
@@ -0,0 +1,108 @@
...
+}
+
+const Onboarding: FC<IOnboarding> = ({ onFinish }) => {
+  const [page, setPage] = useState(0)
+  const pages = 3
+  const next = () => {
+    setPage((page + 1) % (pages))
+  }
+
+  const prev = () => setPage(page - 1)
```
juliangruber, line 47: `next` protects against going over 3, `prev` doesn't. Was this intentional? I think we don't need this protection here inside `next`, and then could also remove the `pages` variable

juliangruber (changes requested):

renderer/src/configs/Routes.tsx
```diff
@@ -0,0 +1,16 @@
...
+const AppRoutes = () => {
+  return (
+    <Router>
+      <Routes>
+        <Route path="/" element={<Welcome />} />
+        <Route path="/dashboard" element={<Saturn />} />
+      </Routes>
+    </Router>
+  )
+}
```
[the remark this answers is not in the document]
jopedroliveira: 9698904768fc81185f2e7451d88c7697f10dbf40

renderer/src/components/InterfaceCalls.tsx
jopedroliveira: I'm not convinced with that location too. 
However, I'm still not sure about the final aspect of this interfaces, and just placed it in the `components` folder. `renderer/src/lib` seems perfect.
Maybe moving all the api calls to a configuration-like object and using it in the frontend, like `config.setOnboarding`.  Nevertheless, I will move it to `lib`

renderer/src/components/Onboarding.tsx
jopedroliveira: 454f03b01a2e33721642f74be50c0b81892df2d5

renderer/src/components/Onboarding.tsx
jopedroliveira: 2e891618b82a949e5177ffcb11afbe6523e8d4c4

renderer/src/components/Onboarding.tsx
jopedroliveira: Similar on Onboarding component: 2e891618b82a949e5177ffcb11afbe6523e8d4c4

renderer/src/components/Onboarding.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  return (
+    <div>
+      {page === 0 &&
```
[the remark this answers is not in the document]
jopedroliveira: aa46b9c9e99be9946f6145f12f03aef1576105c7

renderer/src/components/InterfaceCalls.tsx
juliangruber: @bajtos what do you think?

juliangruber (approved):

renderer/src/components/InterfaceCalls.tsx
juliangruber: sorry, forgot about this unresolved comment

juliangruber (changes requested):

renderer/src/components/InterfaceCalls.tsx
jopedroliveira: Sure, b92c4873ce82e1851db6bfc4c6a9867ad425f5ad

juliangruber (approved):
good stuff

This is great stuff!!

renderer/src/test/onboarding.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  return {
+    setOnboardingCompleted: () => Promise.resolve(undefined),
+    getOnboardingCompleted: (status: boolean) => Promise.resolve(false)
+  }
+})
+
+const mockedUsedNavigate = vi.fn()
+
+vi.mock('react-router-dom', async () => {
+  const router:any = await vi.importActual('react-router-dom')
```
bajtos, line 17: It would be great if we could use the actual type of `import('react-router-dom')` instead of `any`.

Would something like this work?

```suggestion
  const router: import('react-router-dom') = await vi.importActual('react-router-dom')
```

Let's leave this for later, after this PR is landed.

test/e2e/app-launch.e2e.test.js
```diff
@@ -70,7 +70,7 @@ test.describe.serial('Application launch', async () => {
     await mainWindow.click('button:has-text("Continue")')
     await mainWindow.click('button:has-text("Continue")')
     await mainWindow.click('button:has-text("Accept")')
-    expect(await mainWindow.title()).toBe('Saturn')
+    expect(await mainWindow.title()).toBe('Filecoin Station')
```
bajtos, line 73: This does not seem like the right fix to me. I think you should change the steps above to click on the button leading to the Saturn page and then keep this `expect()` unchanged.

juliangruber: right, the test wants to navigate to Saturn 👍

bajtos: Maybe there should also be a test called "navigate through the onboarding flow"? 

My intention behind these end-to-end tests was to verify that we can render all pages ("Station" and "Saturn" at that time), thus preventing trivial errors.

I think this is still useful, and therefore it would be good to update our end-to-end test suite to keep visiting all pages.

At the same time, I am fine to remove tests and pages that are no longer useful.

bajtos: To expedite the landing of this large PR, I am proposing to keep this test as is but change the name from "navigate to Saturn" to something like "complete the onboarding flow" or a different name matching the actual test implementation.

bajtos: Your fix in [067c2de](https://github.com/filecoin-project/filecoin-station/pull/140/commits/067c2de65341a43f9967761f78e212b325f157ee) is good too 👍🏻

juliangruber:
🥳

bajtos:
🚀

PR #142, opened 2022-09-22
Add telemetry. Closes #79

### LINKED ISSUE #79: Telemetry
We want to collect metrics about the stations.

- [ ] Setup infra for collecting metrics (prior art: IPFS Desktop)
- Inside the Station, report the following metrics:
  - [ ] Number of jobs processed per hour
  - [ ] Number of stations up & running
     - Maybe we can report two data points - whether the station is running and whether Saturn L2 is connected to the network.
     - Do we want to report the Station version?
     - Do we want to report whether we are connected to Saturn test-net or main-net?
- [ ] Add a new paragraph to the Consent screen to describe what telemetry data is collected. If the consent screen is not implemented yet, then add a new subtask to the appropriate issue.
- [ ] Setup a simple dashboard to visualise collected data
- [ ] Create follow-up issues for anything left out of scope

Other metrics we can consider - but it's not clear whether users will be happy to share them and how much value we can actually get from them:
- Machine spec: CPU power, Memory size, Storage size
- Network speed test against an L1 node

juliangruber:
Closes #79


main/job-stats.js
```diff
@@ -1,7 +1,23 @@
...
+const writeClient = client.getWriteApi(
+  'julian.gruber@protocol.ai',
+  'station',
+  'ns'
+)
+
+setInterval(() => {
+  writeClient.flush()
+}, 1000)
+
```
bajtos, line 20: Let's move the integration with InfluxDB into a standalone file so that we can easily report telemetry from other places too. How about `main/telemetry-client.js` exporting `writeClient`?

main/job-stats.js
```diff
@@ -1,7 +1,23 @@
...
+const client = new InfluxDB({
+  url: 'https://eu-central-1-1.aws.cloud2.influxdata.com',
+  // station-anonymous-write
+  token: 'P9INe7-DkMB_oYSAXgw0IlBcqRc4bDi05jZFz0NZd8aApv5lieqBiN59M8cmHs6yNMRtly8_KD9RYxnyrnx8dA=='
+})
+const writeClient = client.getWriteApi(
+  'julian.gruber@protocol.ai',
+  'station',
+  'ns'
+)
```
bajtos, line 15: Where are these "magic string" values coming from, and what is their meaning? 

Is it a good idea to use your email address? Shouldn't we use some alias instead, e.g. `station@protocol.ai` or `station@filecoin.io`?

bajtos (commented):
Nice start!

Do we want to attach a unique id of the Station installation to every point we report to InfluxDB? So that we can calculate values per station instance.

How can we test this feature? Ideally, we should have a unit/integration test that calls `JobStats` API and verifies that telemetry events were recorded. 

To me, there are two important parts: 
1. Ensure our internal design & architecture allows us to write such tests.
2. Create the necessary testing infrastructure so that it's very easy to add new tests in the future, e.g. to verify a bug fix.

main/job-stats.js
juliangruber: I'm still figuring things out :) I agree we should use different credentials. The second value is the bucket name, the third I don't know yet (copied this over from their example). Will annotate these 👍

juliangruber:
I'm not sure if Influx is good at handling counters: https://community.influxdata.com/t/handling-counter-data-in-influxdb-2/21596

It seems to be good at answering "how many jobs were completed in the current time frame", but if you for example look at the data for the last 5 minutes, it will always start at 0, since it only counts up what's in the displayed. If we move off the free Influx plan then we get unlimited data retention, so we _should_ always be able to see the total count, but again if we look at 15m intervals for example we will only see the relative increase.

We could change Station to send the current absolute values, but then we also need to attach unique Station installation IDs. Otherwise, if we see two data points of 8000, we can't tell if this is one Station reporting twice that it completed 8000 jobs, or two Stations, having completed 16000 jobs in total. Attaching a station ID (could also be a hash of the wallet address) comes with privacy downsides, but I think acceptable ones. I'll check if Influx works well with this different data model.

bajtos:
> It seems to be good at answering "how many jobs were completed in the current time frame", but if you for example look at the data for the last 5 minutes, it will always start at 0, since it only counts up what's in the displayed. If we move off the free Influx plan, we get unlimited data retention, so we should always be able to see the total count, but again if we look at 15m intervals, for example, we will only see the relative increase.

As I was thinking about this, maybe "jobs completed in the current time frame" is a more useful metric than "total jobs completed since ever"?

Compare:
- We completed 150k jobs yesterday but only 100k jobs the day before.
- We completed 1165k jobs until yesterday but only 1155k jobs until the day before.

But also:
- The swarm of Filecoin Stations handles 2.5k jobs per minute
- The swarm of Filecoin Stations has handled 1165012 jobs so far

I see value in both metrics.

> We could change Station to send the current absolute values, but then we also need to attach unique Station installation IDs. (...) Attaching a station ID (could also be a hash of the wallet address) comes with privacy downsides, but I think acceptable ones. Do we have to comply with GDPR then? I'll check if Influx works well with this different data model.

I think we stay GDPR-compliant as long as the station ID is not linked to any PII (personally identifiable information). In other words, it must be impossible to use station ID to identify the user running the station.

To further improve the privacy of our users, maybe we can periodically rotate the station id, e.g. every day. That way, any information stored about the behaviour of a single station covers only a short time (e.g. one day). However, I am unsure how such an ID rotation would affect our algorithm calculating the total number of jobs processed.

juliangruber:
Get distinct station ids:

```influx
from(bucket: "station")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_measurement"] == "jobs-completed" and r._field == "station")
  |> keep(columns: ["_value"])
  |> sort(columns: ["_value"])
  |> distinct()
  |> count()
```

juliangruber:
Screenshot 2022-09-30 at 12 15 09

main/ui.js
```diff
@@ -24,8 +23,8 @@ module.exports = async function (ctx) {
   const ui = new BrowserWindow({
     title: 'Filecoin Station',
     show: false, // we show it via ready-to-show
-    width: store.get('ui.width', dimensions.width < 1440 ? dimensions.width : 1440),
-    height: store.get('ui.height', dimensions.height < 900 ? dimensions.height : 900),
```
juliangruber, line 28: I removed these since nothing was writing `ui.height`, and simplified the logic

main/store.js
```diff
@@ -1,21 +1,27 @@
 'use strict'
 
 const Store = require('electron-store')
-const electron = require('electron')
+const { randomUUID } = require('crypto')
 
-const defaults = {
-  // https://github.com/sindresorhus/electron-store#defaults
-  language: (electron.app).getLocale() // useful for i18n
-}
```
juliangruber, line 9: I cleaned up the unused keys

main/store.js
```diff
@@ -1,21 +1,27 @@
...
+  defaults: {
+    stationID: randomUUID()
+  }
+})
 
-/** @type {import('electron-store').Options<{language: string}>["migrations"]} */
-const migrations = {
-  // https://github.com/sindresorhus/electron-store#migrations
+const keys = {
+  stationID: 'stationID'
```
juliangruber, line 13: in contrast to other store configs, I decided not to start variables with an uppercase letter, since those are usually reserved for classes and constants. Wdyt, @bajtos?

juliangruber:
```
TypeError: Cannot read properties of undefined (reading 'getPath')
```

Looks like we have a circular dependency now, with `app` not being available 🤔

main/store.js
juliangruber: wouldn't constants be all upper case `STATION_ID` more typically? I'm fine with either as long as it follows a common convention

bajtos: Yeah, all upper case is a common pattern for naming constants.

Recently, I read somewhere an opinion advocating for using PascalCase for constants and the arguments made sense, so I wanted to try it out in practice. I quite like how the code reads this way, but it's subjective.

I don't mind either way: `stationID`, `STATION_ID`, `StationID` or `StationId`. (I slightly prefer `Id` over `ID` but not a big deal.)

I think it would be good to be consistent. Agree on one style, commit to the decision and move on. If possible, configure a linter rule to enforce the style.

juliangruber: I'm fine with either I think, PascalCase is easier to read, but I only want to go there if this either is a common pattern in the ecosystem, or really is a significant improvement. Otherwise I think the "differentness" isn't worth it. This is a very common argument for me, what do you think about this?

Also +1 to enforcing this through lint rules :)

main/ui.js
juliangruber: No I totally agree, PRed [here](https://github.com/filecoin-station/filecoin-station/pull/238)

main/saturn-node.js
```diff
@@ -9,8 +8,7 @@ const { fetch } = require('undici')
 const fs = require('node:fs/promises')
 const path = require('path')
 const { setTimeout } = require('timers/promises')
-
-const configStore = new Store()
+const { getFilAddress, setFilAddress } = require('./station-config')
```
bajtos, line 11: Is this backwards compatible? When I upgrade my current Station that already has my FIL address configured, will the new version find that FIL address?

main/saturn-node.js
```diff
@@ -30,12 +28,6 @@ let webUrl
 /** @type {string | undefined} */
 let apiUrl
 
-const ConfigKeys = {
-  FilAddress: 'saturn.filAddress'
```
bajtos, line 34: Here we use the key `saturn.filAddress`, but the new code uses the key `station.filAddress`. I think that will break compatibility.

Maybe we can use electron-store migrations to "upgrade" the config from the key name to the new key?
https://www.npmjs.com/package/electron-store#migrations

Alternatively, we can implement a fallback in `getFilAddress` to check the old key if there is no value stored under the new key.

What do you think?

Considering how few users we have right now, I can live with breaking the compatibility too. Having said that, I think this is an excellent opportunity to learn how to handle breaking changes in the future when they become more of a problem.

main/job-stats.js
```diff
@@ -24,6 +29,19 @@ class JobStats {
...
+    const filAddress = getFilAddress()
+    if (filAddress && moduleName in this.#perModuleJobStats) {
+      const diff = count - this.#perModuleJobStats[moduleName]
+      if (diff > 0) {
+        writeClient.writePoint(
+          new Point('jobs-completed')
+            .stringField('module', moduleName)
+            .stringField('wallet', createHash('sha256').update(filAddress).digest('hex'))
+            .stringField('station', getStationID())
+            .intField('value', diff)
```
bajtos, line 41: I find this a bit too much involved with details that are not relevant to reporting the number of jobs completed. I would expect that both wallet and station IDs will be added to all points; they are not specific to job stats. Are they?

Can we move the code setting `wallet` and `station` fields into our `telemetry` helper file? For example:


```js
exports.writePoint = function writePoint(point) {
  point          
    .stringField('wallet', createHash('sha256').update(filAddress).digest('hex'))
    .stringField('station', getStationID())
  writeClient.writePoint(point)
}
```

and example usage in `job-stats.js`:

```suggestion
        writePoint(
          new Point('jobs-completed')
            .stringField('module', moduleName)
            .intField('value', diff)
```

What do you think?

bajtos (commented):
Looks mostly good! 👏🏻

I have a few comments to discuss 👇🏻

main/telemetry.js
```diff
@@ -0,0 +1,23 @@
...
+})
+
+const writeClient = client.getWriteApi(
+  'Filecoin Station', // org
+  'station', // bucket
+  'ns' // precision
+)
+
+setInterval(() => {
+  writeClient.flush()
```
bajtos, line 18: What is the error-handling story for `flush()` - can it ever fail? We don't want to trigger unhandled rejection or uncaught error when the Station cannot reach InfluxDB.

main/saturn-node.js
juliangruber: As long as they keys stay the same, this is backwards compatible. Since the store from `station-config` doesn't have a `.name` configured, and the store from `saturn-node` didn't have one either, they were effectively the same store.

main/saturn-node.js
juliangruber: Great idea! I've implemented migrations, and also made some other keys more consistent

main/job-stats.js
juliangruber: This is a great idea! Done just like that

main/telemetry.js
juliangruber: Great catch, I added a `.catch` https://influxdata.github.io/influxdb-client-js/influxdb-client.writeapi.flush.html

main/telemetry.js
```diff
@@ -15,9 +20,23 @@ const writeClient = client.getWriteApi(
...
 
 setInterval(() => {
-  writeClient.flush()
+  writeClient.flush().catch(console.error)
 }, 5000).unref()
 
+/**
+ * @param {Point} point
+ */
+const writePoint = point => {
```
bajtos, line 29: Is there any particular reason for choosing `=>` over `function` syntax?

```suggestion
function writePoint (point) {
```

I personally prefer `function` over `=>`, but I am happy to reconsider.

main/telemetry.js
```diff
@@ -15,9 +20,23 @@ const writeClient = client.getWriteApi(
 )
 
 setInterval(() => {
-  writeClient.flush()
+  writeClient.flush().catch(console.error)
```
bajtos, line 23: Should we report this error to Sentry as well? See

https://github.com/filecoin-station/filecoin-station/blob/11fdc91470d864faeff5323387942d6d54a3feae/main/index.js#L47

For the longer term, it would be great to have a helper function we can use to handle errors we don't want to expose to the user.

Let's keep your current code so we can land this PR sooner and then refactor it in a follow-up PR.

bajtos (approved):
Awesome! 👏🏻 

Once this is landed, it would be great to add another metric to keep track of how many Stations are running.

juliangruber:
> Once this is landed, it would be great to add another metric to keep track of how many Stations are running.

That's already part of this PR :) The `station` field lets us count distinct station apps.

main/telemetry.js
juliangruber: I personally prefer arrow functions, since they can be shorter/more versatile, and V8 still can tell the function name:

```
> const foo = () => { throw new Error('oops') }
undefined
> foo()
Uncaught Error: oops
    at foo (REPL1:1:27)
```

I'm happy to change this to a `function` if you feel strongly about it.

main/telemetry.js
juliangruber: I think the change was small enough to just add on to this PR 👍 

About the helper function, for now since it's just one line of Sentry I don't feel the need to refactor this.

main/telemetry.js
bajtos: Fair enough 👍🏻

bajtos:
> > Once this is landed, it would be great to add another metric to keep track of how many Stations are running.
> 
> That's already part of this PR :) The `station` field lets us count distinct station apps.

Nice!

IIUC, this allows us to count Station that are actively handling jobs, right? If there is a Station running but does not receive any L2 Saturn jobs (retrieval requests), then it will not show in our dashboard. Is that right? Is it a problem?

juliangruber:
> If there is a Station running but it does not receive any L2 Saturn jobs (retrieval requests), then it will not show in our dashboard. Is that a correct description? Is it a problem?

Good catch! Yes that is what's happening. In order to get the station count independent of jobs received, we need to also start sending heartbeats. I can't tell yet whether this could be a problem 🤔

Issue #118, opened 2022-09-13
Consent with telemetry & error reporting


bajtos:
- [ ] Smallprint on the website before download
- [x] A new screen in the app - see #70
 
Things for which want to obtain user's consent:

  -  [x] Usage analytics #80  #160
  -  [x] Telemetry #79 
  -  [x] UI error monitoring #78 #160

juliangruber:
Removed from the milestone, as the current consent screen already covers telemetry, and the website is outside this milestone

bajtos:
Should we perhaps close this issue as done and open two follow-up issues - one for the website and another for the telemetry?

juliangruber:
Telemetry already is covered by issues, I'll share with Onda that we have this website requirement. I'm not sure yet where issues for the website will be tracked. I think we can close this then!

PR #160, opened 2022-09-27
Add load scripts after consent #118

juliangruber:
Partially handles #118, as we don't have Telemetry yet

### DIFF
--- a/renderer/src/components/Plausible.tsx
+++ b/renderer/src/components/Plausible.tsx
@@ -1,0 +1,17 @@
+import { Helmet, HelmetProvider } from 'react-helmet-async'
+
+const Plausible = () => {
+  return (
+    <HelmetProvider>
+      <Helmet>
+        <script
+          defer
+          data-domain="app.station.filecoin.io"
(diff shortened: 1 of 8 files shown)


renderer/src/components/Plausible.tsx
```diff
@@ -0,0 +1,17 @@
+import { Helmet, HelmetProvider } from 'react-helmet-async'
+
+const Plausible = () => {
+  return (
+    <HelmetProvider>
+      <Helmet>
```
bajtos, line 6: I am not familiar with Helmet, but I find it a bit surprising that we are using both `HelmetProvide` and `Helmet` elements here. Shouldn't we put `<HelmetProvider>` to `App.tsx` to make `Helmet` available to all pages and components?

bajtos (approved):
Besides the comment above, the changes look reasonable to me 👍🏻

renderer/src/components/Plausible.tsx
juliangruber: good point!

PR #161, opened 2022-09-27
Add app menu for Windows. Closes #63

### LINKED ISSUE #63: Add "Check for updates" to Windows menu
follows up https://github.com/filecoin-project/filecoin-station/pull/62#discussion_r931039942

juliangruber:
image

image

### DIFF
--- a/main/ui.js
+++ b/main/ui.js
@@ -22,7 +22,6 @@ module.exports = async function (ctx) {
     height: store.get('ui.height', dimensions.height < 900 ? dimensions.height : 900),
     minWidth: 720,
     minHeight: 540,
-    autoHideMenuBar: true,
     titleBarStyle: 'hiddenInset',
     webPreferences: {
       nodeIntegration: false,
--- a/main/app-menu.js
+++ b/main/app-menu.js
@@ -11,8 +11,7 @@ function setupAppMenu (/** @type {import('./typings').Context} */ ctx) {
 
   setupCheckForUpdatesMenuItem(ctx, menu)
 
-  // File menu
-  menu.items[1].submenu?.insert(1, new MenuItem({
+  const saveSaturnModuleLogAs = new MenuItem({
     label: 'Save Saturn Module Log As…',
     click: async () => {
       const opts = { defaultPath: 'station.txt' }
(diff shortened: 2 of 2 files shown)


main/ui.js
```diff
@@ -22,7 +22,6 @@ module.exports = async function (ctx) {
     height: store.get('ui.height', dimensions.height < 900 ? dimensions.height : 900),
     minWidth: 720,
     minHeight: 540,
-    autoHideMenuBar: true,
```
bajtos, line 25: Are you sure we want to show the menu bar? I think the idea was to fully control Station's window to provide consistent design & interaction patterns and also have the same UI on all platforms.

Mixing Windows-native menubar rendered with built-in Windows fonts and Electron-rendered screens with our custom fonts is going to look ugly IMO.

Maybe this is a hint that we should re-think our current usage of the OS menu bar and implement our custom in-app menu instead.

Anyhow, I am ok to iterate on this. We can land your pull request as it is now.

bajtos (approved):

main/ui.js
juliangruber: I agree this is not a great experience, but we haven't yet designed a better one. Maybe we can figure something out until release, but for now this is will get the functional side of things done. Or we could say that we rather have a cleaner window look and ditch the items "Check for updates" and "Save Saturn Module Log As..." (or we could move them to the tray). What do you think?

bajtos: My preference is to have a cleaner window look. I like to idea of adding "Save Saturn Module Log As..." to the Tray menu. I believe "Check for updates" is already there :)

juliangruber: good point, then I'll move the log item there and disable the menus again 👍

bajtos: See https://github.com/filecoin-project/filecoin-station/blob/bd89b53f91a4cbec4b1c60fd2ef82bc3f3522d89/main/tray.js#L29-L39

I'll leave the decision up to you.

If you decide to show the menu bar, then please create a follow-up GH issue to remind us to revisit this part of the design.

bajtos (approved):

PR #163, opened 2022-09-27
build: create Sentry releases & upload source maps

bajtos:
Add new CI/CD build steps to create Sentry Release entry and upload source maps.

See https://github.com/filecoin-project/filecoin-station/issues/78

Sentry docs: https://docs.sentry.io/product/cli/releases/

Initially, I started with `getsentry/action-release@v1`. Unfortunately, it does not allow us to connect custom version numbers (e.g. `0.6.0.2134`) with git commit SHAs. Thus I reworked our integration to use directly `sentry-cli`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,8 @@ jobs:
 
       - name: Check out Git repository
         uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
 
       - name: Install Node.js
         uses: actions/setup-node@v2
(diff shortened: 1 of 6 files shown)


juliangruber (approved):
Thank you for figuring this out! That's strangely a lot of work for a smooth Sentry integration

bajtos:
> Thank you for figuring this out! That's strangely a lot of work for a smooth Sentry integration

I think our use case is a bit special. 

IIUC, the default Sentry setup assumes commit SHA for dev builds and git tag name for release builds. If we used this setup, then the integration would be much smoother. (But we would need to find a way how to pass commit SHA to our Electron app.)

juliangruber:
> But we would need to find a way how to pass commit SHA to our Electron app

Would a synchronous `git` call on launch make for a simpler integration? 🤔

bajtos:
> > But we would need to find a way how to pass commit SHA to our Electron app
> 
> Would a synchronous `git` call on launch make for a simpler integration? 🤔

Can you execute an external process synchronously as part of the first tick?

Once we disable Sentry reporting in dev mode, there will be an option to rework the way how we are constructing the version number for non-release builds from `x.y.z.${GITHUB_RUN_NUMBER}`  to `${GITHUB_SHA}`. But since we already got this far, I feel it's better to keep the current versioning scheme.

juliangruber:
> Can you execute an external process synchronously as part of the first tick?

Doesn't https://nodejs.org/api/child_process.html#child_processexecsynccommand-options give us this?

bajtos:
> > Can you execute an external process synchronously as part of the first tick?
> 
> Doesn't https://nodejs.org/api/child_process.html#child_processexecsynccommand-options give us this?

Nice! I completely forgot about this one. Back in 2013/14, when I joined Node, `execSync` was considered a significant challenge to implement.

PR #164, opened 2022-09-27
fix remove dynamic window title

juliangruber:
image

Anything but "Filecoin Station" is usually not what you expect from a desktop application.


bajtos (approved):
💯

PR #168, opened 2022-09-27
Add dialog when closing Station for the first time

juliangruber:
Screenshot 2022-09-27 at 19 04 37

Screenshot 2022-09-27 at 19 04 58

Copy will land independently.


bajtos:
I think the UX would be nicer if we were showing dialogs inside the HTML frontend, using the same visual design as the rest of the app. I feel the native OS dialog has too different visual style.

Another thing to consider - can we show a different image from the default Electron icon? I am referring to the macOS screenshot.

juliangruber:
> Another thing to consider - can we show a different image from the default Electron icon? I am referring to the macOS screenshot.

This is always the current app icon, so once it's bundled we see the Filecoin Station logo there :)

PR #170, opened 2022-09-27
add: filAddress check to onboarding

jopedroliveira:
This PRs adds to the Onboarding page (entrypoint) the capacity to route the user to `WalletConfig` or `Dashboard` based on previous wallet configuration. 
if the user has already seen the onboarding and configured the wallet, will be redirected to the dashboard page.

### DIFF
--- a/renderer/src/pages/Onboarding.tsx
+++ b/renderer/src/pages/Onboarding.tsx
@@ -1,6 +1,6 @@
 import { useState, useEffect, useCallback } from 'react'
 import { useNavigate } from 'react-router-dom'
-import { getOnboardingCompleted, setOnboardingCompleted } from '../lib/station-config'
+import { getFilAddress, getOnboardingCompleted, setOnboardingCompleted } from '../lib/station-config'
 import Onboarding from '../components/Onboarding'
 import { ReactComponent as StationLogoLight } from '../assets/img/station-logo-light.svg'
 
@@ -20,20 +20,24 @@ const OnboardingPage = (): JSX.Element => {
   const navigate = useNavigate()
   const [isLoading, setIsLoading] = useState<boolean>(true)
   const [isOnboardingCompleted, setIsOnboardingCompleted] = useState<boolean|null>()
+  const [filAddress, setFilAddress] = useState<string|null>()
 
   useEffect(() => {
     (async () => {
       await sleep(2000)
+      setFilAddress(await getFilAddress())
       setIsOnboardingCompleted(await getOnboardingCompleted())
       setIsLoading(false)
     })()
   }, [])
 
   useEffect(() => {
-    if (isOnboardingCompleted) {
+    if (isOnboardingCompleted && !filAddress) {
       navigate('/wallet', { replace: true })
+    } else if (filAddress) {
+      navigate('/dashboard', { replace: true })
     }
-  }, [isOnboardingCompleted, navigate])
+  }, [filAddress, isOnboardingCompleted, navigate])
 
   const onFinishOnboarding = useCallback(async () => {
     setIsOnboardingCompleted(true)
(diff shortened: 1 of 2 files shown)


juliangruber:
Another way to solve this could be:

1. We make `/` the dashboard page
2. Based on the config loaded, we may redirect to `/onboarding` or `/wallet`

To me this has one main advantage: It keeps data loading where it's relevant. The dashboard needs to load config to operate, when we then there notices something is wrong, we can redirect. It's unintuitive why the onboarding component would need to load the fil address. Once https://reactrouter.com/en/dev/fetch/redirect has landed, we can also move this redirect to the react-router configuration.

juliangruber (changes requested):
see comment above

jopedroliveira:
Routing directly to dashboard seems legit, although I would prefere to keep that logic out of `Dashboard`

There are two ux flow involved here that may be affected:
- splash screen/"welcome to station" screen that currently is the Station logo only
- skeleton loading not yet implemented but part of the design flow for dashboard elements

Regarding those above, I like the idea of having a "weolcome screen". Yet, I don't see a huge need for a loading mechanism such as skeleton loading for now, everything seems to be pretty fast (and it might force us to implement a timer so skeleton loading doesn't get weird). 

That being said, I'm ok refactoring the routing logic. But maybe we should prioritise it

juliangruber:
Ok I definitely see your point, let's get this operational first and if I care enough still I'll try to come up with a PR that we can discuss :)

renderer/src/pages/Onboarding.tsx
```diff
@@ -20,20 +20,24 @@ const OnboardingPage = (): JSX.Element => {
...
 
   useEffect(() => {
-    if (isOnboardingCompleted) {
+    if (isOnboardingCompleted && !filAddress) {
       navigate('/wallet', { replace: true })
+    } else if (filAddress) {
+      navigate('/dashboard', { replace: true })
     }
-  }, [isOnboardingCompleted, navigate])
+  }, [filAddress, isOnboardingCompleted, navigate])
```
juliangruber, line 40: This is a suggestion for lines 23 - 40 (of the new file):

```js
  useEffect(() => {
    (async () => {
      await sleep(2000)
      if (await getFilAddress()) {
        return navigate('/dashboard', { replace: true })
      }
      setIsOnboardingCompleted(await getOnboardingCompleted())
      setIsLoading(false)
    })()
  }, [navigate])

  useEffect(() => {
    if (isOnboardingCompleted) {
      navigate('/wallet', { replace: true })
    } 
  }, [isOnboardingCompleted, navigate])
````

I believe there's no advantage in having `filAddress` as a state value

juliangruber (changes requested):

juliangruber (approved):
+1, once tests pass 🙏

bajtos:
This is beyond my knowledge of React routing, but I quite like the suggestion to have a non-root URL for the onboarding screen. Maybe we can keep the root route `/` for the welcome screen and then have routes for everything else (`/dashboard`, `/onboarding`, `/wallet`). WDYT? I also agree with Julian to focus on fixing the user-facing issues first.

juliangruber:
> This is beyond my knowledge of React routing, but I quite like the suggestion to have a non-root URL for the onboarding screen. Maybe we can keep the root route `/` for the welcome screen and then have routes for everything else (`/dashboard`, `/onboarding`, `/wallet`). WDYT?

This would work for me too 👍

jopedroliveira:
I had to implement a workaround to ensure test's running order is done properly.
Mocking the same objects multiple times under the same `describe` is making some tests fail. Instead, I splitted `onboarding` tests in two files.

I'm not sure what's the correct approach to follow with vitest regarding test structure (under same `describer` or separated in different files). Currently everything seems to be working, but this is something worth visiting in a near future.

PR #174, opened 2022-09-28
Add dialog to confirm changing wallet address. Closes #169

### LINKED ISSUE #169: Add dialog when changing wallet address
Screenshot 2022-09-27 at 19 08 07

<del>Blocked by https://github.com/filecoin-project/filecoin-station/pull/144</del>.

juliangruber:
Closes #169

Screenshot 2022-09-28 at 14 14 08

### DIFF
--- a/main/dialog.js
+++ b/main/dialog.js
@@ -3,8 +3,11 @@
 const { IS_MAC } = require('./consts')
 const { dialog } = require('electron')
 
+/** @typedef {import('./typings').Context} Context */
+
 module.exports = {
-  showDialogSync
+  showDialogSync,
+  setup
 }
 
 /**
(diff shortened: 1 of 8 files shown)


bajtos (approved):
It feels to me that showing dialogs via native OS API requires too much complexity in the IPC setup, plus I have concerns about the current visual style - re-posting from https://github.com/filecoin-project/filecoin-station/pull/168#issuecomment-1262175952:
> I think the UX would be nicer if we were showing dialogs inside the HTML frontend, using the same visual design as the rest of the app. I feel the native OS dialog has too different visual style.
>
> Another thing to consider - can we show a different image from the default Electron icon? I am referring to the macOS screenshot.

Having wrote that, I am fine to land your proposed solution as an incremental improvement to get us faster to the first Alpha release.

juliangruber:
Let's revisit native vs custom dialogs afterwards. You're right that native styles might not be what we want, but they are good enough and let us iterate quickly.

I'm also concerned with the amount of IPC boilerplate this required, let's revisit this 👍

bajtos:
> Let's revisit native vs custom dialogs afterwards. You're right that native styles might not be what we want, but they are good enough and let us iterate quickly.
> 
> I'm also concerned with the amount of IPC boilerplate this required, let's revisit this 👍

Cool. I opened https://github.com/filecoin-project/filecoin-station/issues/181

PR #177, opened 2022-09-28
Tray icon assets

### LINKED ISSUE #166: dynamic tray icon
Assigned to @jopedroliveira for providing the asset, then to @bajtos for hooking it up 🙏 

Screenshot 2022-09-27 at 16 29 47

jopedroliveira:
Closes #166, this PR adds icon assets for on, off, and update.

PR intentionally marked as draft so @bajtos can work from here and hook up the icons.


jopedroliveira:
I'm not sure if the `off-macos.png` should be an asset. Documentation suggests that that gray color should be created based on the template icon, applying some filters to it. 
Nevertheless, added that one and another one with outline only. We can discuss and explore that later

bajtos:
@jopedroliveira thank you for the new tray assets. We can land your pull request and then I can open a new one to integrate the update notification.

I am confused about the icons with `-outline` suffix. When are we supposed to use them?

How should we use `assets/tray/on-white.png`?

jopedroliveira:
> I am confused about the icons with `-outline` suffix. When are we supposed to use them?

It's an alternative suggestion for the "off" state in case grey doesn't work well.  Can be removed if we see it wont be necessary.
 
> How should we use `assets/tray/on-white.png`?

That's the dark-mode variation. However, the implementation actually doesn't require it to be a png. In fact those variations (off mode as well) should be handled with [Template Image](https://github.com/electron/electron/blob/main/docs/api/native-image.md#template-image). Nevertheless, added in case you want to follow a different approach. 

Let me know if we want to remove unnecessary assets

bajtos:
Thank you for the explanation, @jopedroliveira! I am still learning about Electron 🙈

Let's keep all tray asset files for now. We can easily remove them later.

> In fact those variations (off mode as well) should be handled with [Template Image](https://github.com/electron/electron/blob/main/docs/api/native-image.md#template-image).

I read through the docs and it's not clear to me how to use Template Image to provide light/dark mode support. Do you know where I can find more documentation? A step-by-step guide would be ideal.

juliangruber:
image (2)

It would be great to also switch to white/black tray icons on Windows and Linux, due to uncontrollable color contrast. Also assigning to @jopedroliveira for assets 🙏

jopedroliveira:
> image (2)
> 
> It would be great to also switch to white/black tray icons on Windows and Linux, due to uncontrollable color contrast. ~Also assigning to @jopedroliveira for assets 🙏 ~ Would be great if you could help with the assets, @jopedroliveira 🙏

The assets on this PR should do the job.
Switching from light/dark mode should be managed by this as well

> @bajtos
> 
>  toogle.dark.mode.tray.mov 
> `tray.js`
> 
> ```
> module.exports = function (/** @type {import('./typings').Context} */ ctx) {
>   const image = nativeImage.createFromPath(path.join(__dirname, '../assets/tray/on-macos.png'))
>   image.setTemplateImage(true)
>   tray = new Tray(image)
>   const contextMenu = Menu.buildFromTemplate([ ...
> ```
> 
> This only toggles between dark/light mode. Afaics `off` mode is not configured, but the idea would be the same (if the Sation doesn't have a wallet address, then it's `off` and the tray icon would is faded).

I'm committing the snippet so we can test

jopedroliveira:
Other than this, the `update-available` and `off` state icons would be defined when an IPC event is triggered.

jopedroliveira:
@juliangruber could you please test this branch on a windows machine to ensure the blue icon is not used?  🙏

juliangruber:
Since you don't have access to a Windows machine, please tell me if you want me to take a look at something

main/tray.js
```diff
@@ -10,17 +10,10 @@ const path = require('path')
...
-
-function icon (/** @type {'on' | 'off'} */ state) {
-  const dir = path.resolve(path.join(__dirname, '../assets/tray'))
-  if (IS_MAC) return path.join(dir, `${state}-macos.png`)
-  return path.join(dir, `${state}.png`)
-}
-
 module.exports = function (/** @type {import('./typings').Context} */ ctx) {
-  tray = new Tray(icon(on))
+  const image = nativeImage.createFromPath(path.join(__dirname, '../assets/tray/on-macos.png'))
```
bajtos, line 14: I think we need to preserve `icon()` logic that returns different file name for macOS.

bajtos:
I apologise for keeping this thread open for so long.

The next step is to verify that this PR fixes the tray icon to black/white colour depending on light/dark mode. We need to test this change on all platforms (macOS/Linux/Windows) and both with light & dark modes.

We should ignore the alternate icons for now. `update-available` icon will be enabled via #53, and I opened #273 to keep track of implementing the `off` mode.

I am going through the Launchpad program now, so my availability is very limited in the next few weeks. If you, @juliangruber, can take over this pull request and finish the changes, then please do so!

juliangruber:
@bajtos I'll have limited availability because of the Wallet work, let's see who gets to it first 🙏 🏎️

juliangruber:
@jopedroliveira I've updated the assets in this PR according to what I think makes sense for each platform.

There are some assets still missing:

- update=yes on=false os=macOS
- update=yes on=true os=win/lin
- update=yes on=false os=win/lin

Could you please create and add these assets, based on how on/off looks for each OS (as can be seen in this PR)?

jopedroliveira:
> @jopedroliveira I've updated the assets in this PR according to what I think makes sense for each platform.
> 
> There are some assets still missing:
> 
> * update=yes on=false os=macOS
> * update=yes on=true os=win/lin
> * update=yes on=false os=win/lin
> 
> Could you please create and add these assets, based on how on/off looks for each OS (as can be seen in this PR)?

@juliangruber 
Added the icons on the gray version, for the update on [4aaa45a](https://github.com/filecoin-station/filecoin-station/pull/177/commits/4aaa45ae9ebaf75c318d8254f13fa4b3e26c815a)
I noticed that there's an outline version for macOS off state which I think is not used since for macos it would be [changed programmatically](https://github.com/filecoin-station/filecoin-station/pull/177#issuecomment-1290339441). However, I updated it with the gray version on [64cfb35](https://github.com/filecoin-station/filecoin-station/pull/177/commits/64cfb3552c081fe75416031375d5d3d98385d2fc). Let me know if there's anything else you need

PR #180, opened 2022-09-29
Fix start with UI hidden. Closes #179

### LINKED ISSUE #179: Start with UI hidden
> Always start Station with UI hidden, as long as the wallet is configured and Station is ready to go

juliangruber:


main/ui.js
```diff
@@ -65,7 +66,9 @@ module.exports = async function (ctx) {
     if (app.dock) app.dock.show()
     ui.show()
   }
-  ui.once('ready-to-show', ctx.showUI)
+  if (!getFilAddress()) {
+    ui.once('ready-to-show', ctx.showUI)
+  }
```
bajtos, line 71: I feel the intent of this code may not be immediately clear. Would you mind adding a short comment to explain the goal?

```suggestion
  // We want the Station to start hidden in the tray.
  // However, when the user did not complete the onboarding flow,
  // we need to show the app to ask them to finish the setup process
  if (!getFilAddress()) {
    ui.once('ready-to-show', ctx.showUI)
  }
```

Is it enough to check that we have the FIL wallet address configured, or should we also check whether the user has completed the onboarding flow?

juliangruber: > Is it enough to check that we have the FIL wallet address configured, or should we also check whether the user has completed the onboarding flow?

I thought of this too, but don't think there can be a state where onboarding hasn't been completed but a FIL address has been set up 🤔

bajtos (approved):

PR #183, opened 2022-09-29
build: disable Sentry integration for dev builds

bajtos:
This applies to both packaged versions created locally using `npm run package` and to running unpackaged version directly via `npm start`.


renderer/src/components/Sentry.tsx
```diff
@@ -4,6 +4,9 @@ import { BrowserTracing } from '@sentry/tracing'
 
 const SentryComponent = () => {
   useEffect(() => {
+    // Disable Sentry integration for dev builds
+    if (window.electron.stationBuildVersion.match(/\.1-dev$/)) { return }
```
juliangruber, line 8:
```suggestion
    if (window.electron.stationBuildVersion.endsWith('.1-dev')) { return }
```

This would be simpler, but then I wonder if `.1` is a meaningful segment to check for, and maybe we can just have this?

```suggestion
    if (window.electron.stationBuildVersion.endsWith('-dev')) { return }
```

juliangruber (changes requested):

renderer/src/components/Sentry.tsx
bajtos: I agree `endsWith` is a better tool 👍🏻 

`.1` is part of the hard-coded build number `.1-dev`.

At the end of the day, this does not matter much. I am going with your second suggestion.

juliangruber:
too fast! :P

juliangruber (approved):

PR #184, opened 2022-09-29
feat: report unexpected backend errors to Sentry

bajtos:
See #78


main/index.js
```diff
@@ -1,5 +1,7 @@
 'use strict'
 
+require('./setup-sentry-on-backend')
```
juliangruber, line 3:
```suggestion
require('./setup-sentry')
```

"on backend" is given by the location inside the `main` folder

main/index.js
```diff
@@ -82,6 +87,18 @@ const ctx = {
...
+app.on('before-quit', () => {
+  // Flush pending events immediately
+  // See https://docs.sentry.io/platforms/node/configuration/draining/
+  Sentry.close()
+})
+
+process.on('uncaughtException', err => {
+  Sentry.captureException(err)
+  log.error(err)
+  process.exitCode = 1
```
juliangruber, line 99: will this still close the process on uncaught exception?

main/setup-sentry-on-backend.js
```diff
@@ -0,0 +1,20 @@
+'use strict'
+
+const Sentry = require('@sentry/node')
+const { BUILD_VERSION } = require('./consts')
+
+if (BUILD_VERSION.endsWith('-dev')) {
+  // Disable Sentry integration for dev builds
+} else {
```
juliangruber, line 8:
```suggestion
// Disable Sentry integration for dev builds
if (!BUILD_VERSION.endsWith('-dev')) {
```

It's not a logical branch, so I think just a comment is enough

juliangruber (changes requested):

main/index.js
bajtos: > "on backend" is given by the location inside the main folder

You are right.

I am not a fan of having files with names that also require their path to understand the purpose of the file. 

Having said that, I guess `Sentry.tsx` and `setup-sentry.js` are different enough that I can memorise which one is for the frontend and which is for the backend.

main/index.js
bajtos: I am glad you have asked!

I think it depends. When the uncaught exception was created from an unhandled promise rejection, then the process keeps running. If it's an error thrown from a callback, then the app exits.

The current behaviour after an unhandled error/rejection is a bit fuzzy at the moment. I was hoping that my current change was not making the situation any worse, but I could be wrong.

Maybe this is the right time to revisit our error-handling strategy. What would you consider the right solution?

Note that for errors thrown from our setup functions, we don't exit the app.
https://github.com/filecoin-project/filecoin-station/blob/2e38e120688dbbf8f66142f9321d2ac4af61bdfd/main/index.js#L29-L38

main/index.js
juliangruber: I see, depending on the editor this can definitely be a problem! It would be cool if we could find a convention for this 🤔

main/index.js
juliangruber: To me the main concern is for this PR not to change any behavior by accident. If the app was already not quitting on uncaught exceptions, then +1 to keeping this logic, until we decide to change it. If you're confirming that the logic basically is still the same, then I'm +1 with this change

juliangruber (approved):

main/index.js
bajtos: See https://github.com/filecoin-project/filecoin-station/pull/184#issuecomment-1262350907

juliangruber:
<3

PR #197, opened 2022-09-30
Update L2-node to 0.4.2

juliangruber:


build/download-saturn-l2.js
```diff
@@ -31,7 +33,7 @@ async function main () {
   await Promise.all(
     assets
       .map(async ({ name, browser_download_url: url }) => {
-        const match = name.match(/^L2-node_\d+\.\d+\.\d+_([A-Za-z0-9]+)_([A-Za-z0-9_]+)\.tar\.gz$/)
+        const match = name.match(/^L2-node_([A-Za-z0-9]+)_([A-Za-z0-9_]+)\.(tar\.gz|zip)$/)
```
juliangruber, line 39: the version number was removed from the file name, plus apple requires `.zip` for notarization

build/download-saturn-l2.js
```diff
@@ -58,7 +63,27 @@ async function main () {
...
+        if (match[3] === 'tar.gz') {
+          await pipeline(res.body, gunzip(), tar.extract(outDir))
+        } else {
+          // Darwin needs to be a zip for notarization
+          await mkdir(path.join(outDir, 'l2node-darwin-x64'))
+          const parser = unzip.Parse()
+          await Promise.all([
+            (async () => {
+              while (true) {
+                const entry = /** @type UnzipStreamEntry */ (await once(parser, 'entry'))
```
juliangruber, line 75: There's a type error here, do you know how to fix this?

> Conversion of type 'any[]' to type 'Entry' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type 'any[]' is missing the following properties from type 'Entry': path, type, size, autodrain, and 54 more.ts(2352)

I tried

```js
/** @type UnzipStreamEntry */ /** @type unknown **/ (await once(parser, 'entry'))
```

But then there are type issues afterwards

build/download-saturn-l2.js
bajtos: > the version number was removed from the file name,

Hmm, that's not great. I believe it's a best practice to include the version number in the file names of release artefacts.

Is it difficult to fix the L2-node release process to bring back the version number to file names?

I am ok to land what your have now and change this script again later, if/when we add the version number back.

build/download-saturn-l2.js
bajtos: I think you need to put the type name into `{}`.

```suggestion
                const entry = /** @type {UnzipStreamEntry} */ (await once(parser, 'entry'))
```

bajtos (approved):

build/download-saturn-l2.js
juliangruber: The issue is that `gon` doesn't support variables in all fields, so we can't script `zip.output_path`. We could however create a script that renames the artefacts after the fact.

If you look at gon itself for example, they don't have the version number there: https://github.com/mitchellh/gon/releases/tag/v0.2.5.
Goreleaser seems to have assets both with and without version number: https://github.com/goreleaser/goreleaser/releases/tag/v1.11.5, but there's more without so I assume also for them this is the main way to go?

> I believe it's a best practice to include the version number in the file names of release artefacts
I don't think I've heard of that, do you have a source?

bajtos: > The issue is that `gon` doesn't support variables in all fields, so we can't script `zip.output_path`. We could however create a script that renames the artefacts after the fact.

I see. I would consider this topic as "nice to have", nothing major.

> > I believe it's a best practice to include the version number in the file names of release artefacts
> 
> I don't think I've heard of that, do you have a source?

Ha, you got me! Maybe it's just my assumption based on the filenames I have seen on various download pages.

Few examples:

- Node.js: https://nodejs.org/dist/v16.17.1/node-v16.17.1-darwin-x64.tar.gz
- An example Rust project: https://github.com/dandavison/delta/releases/download/0.14.0/delta-0.14.0-x86_64-apple-darwin.tar.gz
- PHP: https://www.php.net/distributions/php-8.1.11.tar.gz
- Git for Windows: https://github.com/git-for-windows/git/releases/download/v2.38.0.windows.1/Git-2.38.0-64-bit.exe

build/download-saturn-l2.js
bajtos: The documentation for `once` in Node typings says this:

> The `Promise` will resolve with an array of all the arguments emitted to the given event.

I added a commit to fix the types and then also another issue I discovered while running the script.

See [2fc3a2a](https://github.com/filecoin-station/filecoin-station/pull/197/commits/2fc3a2a6b8f2ce6220c69c419e763c4b6befdb78)

build/download-saturn-l2.js
```diff
@@ -67,12 +67,12 @@ async function main () {
...
         } else {
           // Darwin needs to be a zip for notarization
-          await mkdir(path.join(outDir, 'l2node-darwin-x64'))
+          await mkdir(path.join(outDir, 'l2node-darwin-x64'), { recursive: true })
           const parser = unzip.Parse()
           await Promise.all([
             (async () => {
               while (true) {
-                const entry = /** @type UnzipStreamEntry */ (await once(parser, 'entry'))
+                const [entry] = /** @type {[UnzipStreamEntry]} */ (await once(parser, 'entry'))
```
juliangruber, line 75: thank you so much for fixing! and weird that it worked before that, I must not have tested properly

PR #199, opened 2022-10-03
Add Update Available banner

jopedroliveira:
Added a new component for the topbar showing information about a new version being available for update.

See https://github.com/filecoin-station/filecoin-station/issues/137

Screenshot 2022-10-03 at 10 04 33


renderer/src/components/Notification.tsx
```diff
@@ -0,0 +1,31 @@
...
+            </div>
+             <button
+              className='py-2 px-8 rounded-full font-body text-body-2xs border border-solid
+              text-white hover:bg-grayscale-100 hover:bg-opacity-30 bg-transparent'
+              onClick={() => callback()}>{ buttonText }</button>
+          </div>
+        </div>
+      }
+    </>
+  )
```
juliangruber, line 28: to reduce indentation

```suggestion
  if (!message) return null
  return (
    <div className="h-14 bg-black top-0 left-0 w-full">
      <div className="h-full max-w-[744px] mx-auto flex flex-row items-center justify-between">
        <div className='flex flex-row items-center gap-1'>
        <Warning width={'12px'} height={'12px'} fill="none" stroke="#fff" />
        <span className='text-white text-body-xs'> { message } </span>
        </div>
         <button
          className='py-2 px-8 rounded-full font-body text-body-2xs border border-solid
          text-white hover:bg-grayscale-100 hover:bg-opacity-30 bg-transparent'
          onClick={() => callback()}>{ buttonText }</button>
      </div>
    </div>
  )
```

juliangruber: @bajtos the linter didn't complain about line 20 being one space too far indented, shall we introduce some more react/html specific rules?

renderer/src/pages/Dashboard.tsx
```diff
@@ -42,16 +44,31 @@ const Dashboard = (): JSX.Element => {
...
+      setNotification(undefined)
+    }
+  }
+
   return (
     <div className="h-screen w-screen overflow-hidden bg-grayscale-100">
+      {notification &&
+        <Notification message={notification.text}
+        buttonText={notification.buttonText}
+        callback={notificationCallback} />
```
juliangruber, line 69: `callback` isn't so semantic, what would be a better prop here? It's not `onDismiss`, it's more like `onClick` however it only fires when you click the button, not the whole notification.

I think we're generalizing this component too much, we don't yet have a concept of in-app notifications, can we refactor it to be specific to this use case? Something like `<UpdateBanner onRestart={}>`. Wdyt?

juliangruber (changes requested):

renderer/src/pages/Dashboard.tsx
jopedroliveira: Seems legit to me.
Hence, should the click `action`  be specific for this component? Calling directly a possible restart API instead of using the `action` property of the object?

juliangruber: Good idea, the component can issue the rpc call itself 👍

renderer/src/components/Notification.tsx
bajtos: > the linter didn't complain about line 20 being one space too far indented, shall we introduce some more react/html specific rules?

That's a good idea! 💯

main/updater.js
```diff
@@ -86,6 +94,7 @@ function onUpdaterError (err) {
  * @param {import('electron-updater').UpdateInfo} info
  */
 function onUpdateAvailable ({ version /*, releaseNotes */ }) {
+  ipcMain.emit(ipcMainEvents.UPDATE_AVAILABLE)
```
bajtos, line 97: Based on the code below, I think we should show emit UPDATE_AVAILABLE event only after we have successfully downloaded the new version.

Since the app will display a banner at the top, I think we should also stop calling `showDialogSync`.

Also: the dialog has two buttons - Close and Show Release Notes. For the in-app notification, we arse showing Restart instead of Close, which is fine. 

Can we please keep the UX for showing release notes? For example, the in-app notification can show two buttons - "Restart" and "Show Release Notes".

main/index.js
```diff
@@ -86,7 +86,8 @@ const ctx = {
   saveSaturnModuleLogAs: () => { throw new Error('never get here') },
   showUI: () => { throw new Error('never get here') },
   loadWebUIFromDist: serve({ directory: path.resolve(__dirname, '../renderer/dist') }),
-  confirmChangeWalletAddress: () => { throw new Error('never get here') }
+  confirmChangeWalletAddress: () => { throw new Error('never get here') },
+  restart: () => { throw new Error('never get here') }
```
bajtos, line 90: This is not a generic restart, the implementation calls `autoUpdater.quitAndInstall()` under the hood.

I propose renaming this new IPC method to `quitAndInstall` or something similar (`restartToUpdate`?) to better communicate that this API is for applying updates only.

What do you think?

bajtos (changes requested):
Nice job, @jopedroliveira! It great to see that you are getting very familiar with the IPC mechanism and building better understanding of the backend code too 👏🏻 

I have two comments to discuss, PTAL 👇🏻

main/updater.js
bajtos: I'd say pick the option that's easiest to implement. Opening the release notes in the default web browser is good enough, IMO.

main/index.js
jopedroliveira: `restartToUpdate` sounds good

juliangruber: good idea!

main/updater.js
jopedroliveira: I added a link for release notes on [900758d](https://github.com/filecoin-station/filecoin-station/pull/199/commits/900758d4d5d42c44576bb376a9981cd4251f5aeb)

Since we're on an electron window, I implemented an RPC so electron can trigger the OS default browser window. The frontend is not able to do that, at least in a good-looking way

jopedroliveira: Screenshot 2022-10-11 at 12 02 50

bajtos (commented):
The link for release notes looks nice 👍🏻

main/updater.js
bajtos: > Since the app will display a banner at the top, I think we should also stop calling `showDialogSync`.

I was looking into the code. We are calling `showDialogSync` only when the update check was explicitly initiated by the user. On the second thought, I think we should preserve this behaviour, to let the user know their requested operation is done.

Let's close this thread as resolved.

main/ipc.js
```diff
@@ -12,7 +12,8 @@ const ipcMainEvents = Object.freeze({
   JOB_STATS_UPDATED: 'station:job-stats-updated',
 
   UPDATE_CHECK_STARTED: 'station:update-check:started',
-  UPDATE_CHECK_FINISHED: 'station:update-check:finished'
+  UPDATE_CHECK_FINISHED: 'station:update-check:finished',
+  UPDATE_AVAILABLE: 'station:update-available'
```
bajtos, line 16: I am afraid it's not enough to emit an IPC event.

- The Station starts running in the Tray only with the main window hidden.
- When the main window is hidden, it does not receive events.
- As a result, the new banner is not shown.

I am not sure why our frontend code is not receiving the event. I believe we do forward IPC events to the window. Maybe when the window was not shown then our frontend code installing event handlers did not run?

I think we need to add a new method allowing the frontend to query the status of the update check.

I am going to look into this myself right now and let you know what I find.

renderer/src/lib/station-config.tsx
```diff
@@ -55,3 +55,7 @@ export async function getTotalJobsCompleted (): Promise<number> {
 export async function restartToUpdate (): Promise<void> {
   return await window.electron.restartToUpdate()
 }
+
+export function openReleaseNotes (): void {
+  return window.electron.stationEvents.openReleaseNotes()
```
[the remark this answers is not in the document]
bajtos: Done in https://github.com/filecoin-station/filecoin-station/pull/199/commits/8de84e0227b23d525e8f7deb2a5ba21c1aa87ce6

main/ipc.js
bajtos: Done in 8de84e0227b23d525e8f7deb2a5ba21c1aa87ce6

bajtos (approved):
With the tweaks I made in the last commit, I think this PR is good to get landed.

End-to-end tests are failing locally for me, but that's unrelated to the changes made here. I'll open a new PR to fix our E2E setup.

main/ipc.js
juliangruber: Interesting, let's keep an eye on this issue in the future 🤔

main/preload.js
```diff
@@ -46,6 +50,12 @@ contextBridge.exposeInMainWorld('electron', {
 
       ipcRenderer.on('station:earnings-counter', listener)
       return () => ipcRenderer.removeListener('station:earnings-counter', callback)
+    },
+    onUpdateAvailable: (/** @type {() => void} */ callback) => {
+      const listener = () => callback()
+
+      ipcRenderer.on('station:update-available', listener)
+      return () => ipcRenderer.removeListener('station:update-available', callback)
```
juliangruber, line 58:
```suggestion
      return () => ipcRenderer.removeListener('station:update-available', listener)
```

We have made this mistake above as well, can you fix them as part of this PR too?

main/updater.js
```diff
@@ -31,7 +41,13 @@ function setup (/** @type {import('./typings').Context} */ _ctx) {
   require('electron').autoUpdater.on('before-quit-for-update', beforeQuitCleanup)
 }
 
-module.exports = async function (/** @type {import('./typings').Context} */ ctx) {
+module.exports = async function setupUpdater (/** @type {import('./typings').Context} */ ctx) {
+  ctx.getUpdaterStatus = async function getUpdaterStatus () {
```
juliangruber, line 45: Does this function need to be `async`?

main/updater.js
```diff
@@ -104,10 +138,15 @@ function onUpdateAvailable ({ version /*, releaseNotes */ }) {
...
 
   if (buttonIx === 1) {
-    shell.openExternal(`https://github.com/filecoin-project/filecoin-station/releases/v${version}`)
+    openReleaseNotes()
   }
 }
 
+function openReleaseNotes () {
+  const version = nextVersion ? `v${nextVersion}` : 'latest'
+  shell.openExternal(`https://github.com/filecoin-project/filecoin-station/releases/${version}`)
```
juliangruber, line 147:
```suggestion
  shell.openExternal(`https://github.com/filecoin-station/filecoin-station/releases/${version}`)
```

renderer/src/components/UpdateBanner.tsx
```diff
@@ -0,0 +1,46 @@
+import { useEffect, useState } from 'react'
+import { ReactComponent as Warning } from '../assets/img/error.svg'
+import { openReleaseNotes, restartToUpdate } from '../lib/station-config'
+
+const UpdateBanner = () => {
+  const [isUpdateAvailable, setIsUpdateAvailable] = useState<boolean>(false)
+
+  useEffect(() => {
+    window.electron.getUpdaterStatus()
+      .then(({ updateAvailable }) => setIsUpdateAvailable(updateAvailable))
```
juliangruber, line 10: can we please again use `async/await`?

juliangruber (changes requested):

main/preload.js
jopedroliveira: [5c37fd3](https://github.com/filecoin-station/filecoin-station/pull/199/commits/5c37fd3fd9ae539d4890239b888c6e83928fcbe4)

main/updater.js
jopedroliveira: [8fbf35e](https://github.com/filecoin-station/filecoin-station/pull/199/commits/8fbf35ebeca5e6657f754767cd71ea7ab2bea294)

renderer/src/components/UpdateBanner.tsx
jopedroliveira: [ba42f5d](https://github.com/filecoin-station/filecoin-station/pull/199/commits/ba42f5d4b1b6fcb8c767d54a77bf71842a7eef8e)

main/updater.js
jopedroliveira: @bajtos

bajtos: I was thinking about making the function future-proof. I am fine with changing it to a regular (sync) function.

renderer/src/components/UpdateBanner.tsx
```diff
@@ -5,12 +5,13 @@ import { openReleaseNotes, restartToUpdate } from '../lib/station-config'
...
-    window.electron.getUpdaterStatus()
-      .then(({ updateAvailable }) => setIsUpdateAvailable(updateAvailable))
-  })
+  const reload = async () => {
+    const updateStatus = await window.electron.getUpdaterStatus()
+    setIsUpdateAvailable(updateStatus.updateAvailable)
+  }
 
   useEffect(() => {
+    reload()
```
bajtos, line 14: I am not a React expert, but their docs explicitly warns against defining async functions outside of `useEffect`.

Quoting from https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies
> **Is it safe to omit functions from the list of dependencies?**
> Generally speaking, no.
> (...)
> It’s difficult to remember which props or state are used by functions outside of the effect. This is why usually you’ll want to declare functions needed by an effect inside of it. Then it’s easy to see what values from the component scope that effect depends on.

jopedroliveira: You're probably right. I followed a pattern we're using somewhere else. Maybe I should refactor it as well in a follow-up PR!

jopedroliveira: 20c7fb2

main/updater.js
bajtos: Fixed in 6dffefc

renderer/src/components/UpdateBanner.tsx
```diff
@@ -5,19 +5,19 @@ import { openReleaseNotes, restartToUpdate } from '../lib/station-config'
...
+    }
+
     reload()
     const unsubscribeUpdateNotification = window.electron.stationEvents.onUpdateAvailable(() => {
       setIsUpdateAvailable(true)
     })
 
     return unsubscribeUpdateNotification
-  }, [])
+  }, [isUpdateAvailable])
```
bajtos, line 20: We are not using `isUpdateAvailable` anywhere inside the effect, are we?

We use `setIsUpdateAvailable` only, which we don't need to list in the dependencies, as state setters are guaranteed never to change. (In React parlance:
The identity of the `setIsUpdateAvailable` function is guaranteed to be stable so it’s safe to omit.)

jopedroliveira: 5cf1f28

juliangruber (approved):

PR #201, opened 2022-10-03
refactor: page routing

### LINKED ISSUE #200: UI routing logic
Following the conversation on #170 

> This is beyond my knowledge of React routing, but I quite like the suggestion to have a non-root URL for the onboarding screen. Maybe we can keep the root route / for the welcome screen and then have routes for everything else (/dashboard, /onboarding, /wallet). WDYT?

Implement routing logic:

`/` Welcome page with the welcome screen and routing logic
`/onboarding` Onboarding page
`/wallet`Wallet input (login) page
`/dashboard` Dashboard page

jopedroliveira:
Closes #200
This PR moves the routing logic for a new page `Welcome`, which displays the welcome screen, then routes to the proper screen depending if the user has completed the onboarding previously and set up a wallet.


renderer/src/test/welcome_onboardingCompleted.test copy.tsx
```diff
@@ -0,0 +1,47 @@
+import { beforeEach, describe, expect, test, vi } from 'vitest'
```
juliangruber, line 1: this test file has an unexpected name, is the ` copy` intentional?

renderer/src/pages/Welcome.tsx
```diff
@@ -0,0 +1,33 @@
...
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
+
+const Welcome = (): JSX.Element => {
+  const navigate = useNavigate()
+
+  useEffect(() => {
+    (async () => {
+      await sleep(2000)
+      if (await getFilAddress()) {
+        return navigate('/dashboard', { replace: true })
```
juliangruber, line 15:
```suggestion
        return navigate('/dashboard')
```

I had a realization that we don't need any of these, since desktop apps usually have no concept of going back in time anyway (the user wouldn't be able to)

renderer/src/pages/Welcome.tsx
```diff
@@ -0,0 +1,33 @@
+import { useEffect } from 'react'
+import { useNavigate } from 'react-router-dom'
+import { getFilAddress, getOnboardingCompleted } from '../lib/station-config'
+import { ReactComponent as StationLogoLight } from '../assets/img/station-logo-light.svg'
+
+const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
+
+const Welcome = (): JSX.Element => {
```
juliangruber, line 8: I like this component for its simplicity and focus, but wonder if we can find a better name. "Welcome" is the purpose to the user, but to the codebase it serves as something else: It checks whether the user as completed onboarding and set up a fil address, and redirects based on those results. Maybe "Redirector" works, it's boring but explicit. I still think overall this isn't the way to go, since now this component needs to know about the requirements of other components. I'm going to create an alternative PR so we can compare solutions :)

juliangruber (changes requested):

renderer/src/test/welcome_onboardingCompleted.test copy.tsx
jopedroliveira: nop

renderer/src/pages/Welcome.tsx
jopedroliveira: Good catch. 
I'll open a follow-up PR to review it on other pages as well.

renderer/src/pages/Welcome.tsx
jopedroliveira: Right. 
I think this will depend on the way the app evolves. In the meantime, I renamed it to make it more explicit as you suggested.

PR #205, opened 2022-10-04
Reduce saturn stats poll frequency for perf. Closes #198

### LINKED ISSUE #198: L2-node CPU usage high
```bash
FIL_WALLET_ADDRESS=... ./build/saturn/l2node-darwin-x64/saturn-L2-node
```

low:
Screenshot 2022-09-30 at 13 47 08

high:
Screenshot 2022-09-30 at 13 48 17

```bash
npm start
```

low:
Screenshot 2022-09-30 at 13 49 44

high:
Screenshot 2022-09-30 at 13 49 33

Why does it consume so much more CPU when spawned by electron? @bajtos do you have an idea?

juliangruber:
Closes #198

One could even argue that a 1s interval is better for the user, because it makes the UI less hectic

### DIFF
--- a/main/saturn-node.js
+++ b/main/saturn-node.js
@@ -137,7 +137,7 @@ async function start (/** @type {Context} */ ctx) {
         pollStatsInterval = setInterval(() => {
           pollStats(ctx)
             .catch(err => console.warn('Cannot fetch Saturn module stats.', err))
-        }, 100)
+        }, 1000)
         resolve()
       }
     }


juliangruber:
@bajtos hindsight please 🙏

bajtos:
I used the 100ms refresh rate to make the counter feel more alive. I think we will need multiple iterations to find the right balance between interactivity and CPU usage. 

I am fine to use 1s refresh rate 👍🏻

juliangruber:
I agree that the previous interval felt more alive 👍  We could also investigate why 10ops/s caused high CPU usage in the first place.

PR #208, opened 2022-10-04
Refactor react pages tests

jopedroliveira:
After some research and explorations, figured out that using the `describe` to define a suit of tests assumes the context is shared by them all.

We are using `describe` for grouping the tests by page, which is fine supposed, however, we were defining each test suit using a nested `describe` which assumes a shared context. 

To avoid context sharing, we replaced the low-level `describe` with `test` to ensure the context is only affecting the contained tests.


juliangruber (approved):

bajtos (commented):
I'll trust your judgment here.

juliangruber (approved):

PR #212, opened 2022-10-05
Bring back Sentry Releases

bajtos:
- Revert "temporarily disable step "Create Sentry Release""
- build: do not create Sentry Releases from forks

Quoting from https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow:

> Note: With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

See #78

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,12 @@ jobs:
           node ./build/configure-sentry.js >> $GITHUB_ENV
 
       - name: Create Sentry release
-        if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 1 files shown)


bajtos:
I am not sure if errors reported by CI tests from a forked repo will be correctly associated with the `development` environment in Sentry. I am proposing to deal with that later, after this PR is landed.

bajtos:
I am not sure if errors reported by CI tests from a forked repo will be correctly associated with the `development` environment in Sentry. I am proposing to deal with that later, after this PR is landed.

PR #213, opened 2022-10-05
build: cancel CI builds after a PR was updated

bajtos:
Configure GitHub Actions Workflow to cancel any pending or in-progress runs of our CI build after the pull request being tested has been modified. We are interested in the result of the new build that will be started for the updated version.

Preserve the old behavior for the `main` branch, where we want to build every commit even when new commits were added before the CI finished.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,12 @@ on:
     tags:
       - '*'
 
+# This cancels in-progress jobs or runs on pull_request events only
+# See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 env:
   ELECTRON_CACHE: ${{ github.workspace }}/.cache/electron
   ELECTRON_BUILDER_CACHE: ${{ github.workspace }}/.cache/electron-builder


juliangruber:
What are you adding this for?

While figuring things out I do find it convenient to build more than one commit.

bajtos:
> What are you adding this for?

When there are multiple pull requests approved & ready to be merged at the same time, we have to land them one by one. Rebase the PR on top of the latest main, wait for CI to complete, land, then rinse & repeat for the next PR.

In my experience, sometimes there is a kind of a race between pull requests being rebased, e.g. when both you and I are trying to get our changes landed.

In such cases, we often rebase the PR on top of the new main branch before the CI build is finished. The CI build for the new version gets delayed because there are CI builds for old versions of the PR waiting in the queue.

My PR is changing that behaviour: if you push new changes to the PR, e.g. after rebasing on top of the new main branch, the old CI build is cancelled & discarded and a new build is started (soon-ish) for the new version.

> While figuring things out I do find it convenient to build more than one commit.

Interesting! In that case, I think we should preserve the current behaviour and close my PR as rejected. WDYT?

juliangruber:
> The CI build for the new version gets delayed because there are CI builds for old versions of the PR waiting in the queue.

I wonder, Is this actually happening? I know with macOS there can be queues sometimes, but are these because we're already building, or because there's a general queue on the machine type?

PR #214, opened 2022-10-07
Fix saturn log buffer overflow. Closes #211

### LINKED ISSUE #211: saturn SIGPIPE
My logs ended with

```
[3:22:33 PM] 2022-10-05T15:22:33.882+0200	INFO	saturn-l2	logs/logs.go:35	car transfer successful	{"id": "c3a11688-6392-3db5-9df3-d9fef1948033"}
[3:22:34 PM] 2022-10-05T15:22:34.299+0200	INFO	l1-interop	l1interop/l1sseclient.go:184	received request from L1	{"l1": "192.240.99.20", "json": "{\"requestId\":\"d67e8edebb42a539e6ffec908709709a\",\"root\":\"bafkreihg6lf3z7mfm6mho5nmvkc55srfgxg7z5rtuixmjowg4y7orzrdqa\"}"}
[3:22:34 PM] 2022-10-05T15:22:34.300+0200	INFO	saturn-l2	logs/logs.go:35	parsed CAR transfer request received from L1	{"id": "d67e8ede-bb42-a539-e6ff-ec908709709a", "l1": "192.240.99.20", "req": {"RequestId":"d67e8ede-bb42-a539-e6ff-ec908709709a","Root":{"/":"bafkreihg6lf3z7mfm6mho5nmvkc55srfgxg7z5rtuixmjowg4y7orzrdqa"},"SkipOffset":0,"Selector":{}}}
[3:22:34 PM] Saturn node exited via signal SIGPIPE
```

The log file is 100MB in size. It looks like an internal buffer overflowed.

juliangruber:
Closes #211

### DIFF
--- a/main/saturn-node.js
+++ b/main/saturn-node.js
@@ -241,9 +241,7 @@ function appendToChildLog (text) {
     .split(/\n/g)
     .map(line => `[${new Date().toLocaleTimeString()}] ${line}`)
   )
-  while (childLog.length > 100) {
-    childLog.shift()
-  }
+  childLog.splice(0, childLog.length - 100)
 }
 
 /**
--- a/main/saturn-node.js
+++ b/main/saturn-node.js
@@ -21,8 +21,8 @@ let childProcess = null
 
 let ready = false
 
-/** @type {string} */
-let childLog = ''
+/** @type {string[]} */
+let childLog = []
 
 /** @type {string | undefined} */
 let webUrl
@@ -86,7 +86,7 @@ async function start (/** @type {Context} */ ctx) {
     return
   }
 
-  childLog = ''
+  childLog = []
   appendToChildLog('Starting Saturn node')
   childProcess = execa(saturnBinaryPath, {
     env: {
@@ -203,7 +203,7 @@ function getWebUrl () {
 }
 
 function getLog () {
-  return childLog
+  return childLog.join('\n')
 }
 
 /**
@@ -236,11 +236,14 @@ function forwardChunkFromSaturn (chunk, log) {
  * @param {string} text
  */
 function appendToChildLog (text) {
-  childLog += text
+  childLog.push(...text
     .trimEnd()
     .split(/\n/g)
-    .map(line => `[${new Date().toLocaleTimeString()}] ${line}\n`)
-    .join('')
+    .map(line => `[${new Date().toLocaleTimeString()}] ${line}`)
+  )
+  while (childLog.length > 100) {
+    childLog.shift()
+  }
 }
 
 /**


main/saturn-node.js
```diff
@@ -86,7 +86,7 @@ async function start (/** @type {Context} */ ctx) {
     return
   }
 
-  childLog = ''
+  childLog = []
```
juliangruber, line 89: converted to an array of lines, to get easier line based operations

bajtos: Yes please, I would do the same 👏🏻

main/saturn-node.js
```diff
@@ -236,11 +236,14 @@ function forwardChunkFromSaturn (chunk, log) {
...
+  childLog.push(...text
     .trimEnd()
     .split(/\n/g)
-    .map(line => `[${new Date().toLocaleTimeString()}] ${line}\n`)
-    .join('')
+    .map(line => `[${new Date().toLocaleTimeString()}] ${line}`)
+  )
+  while (childLog.length > 100) {
+    childLog.shift()
+  }
```
bajtos, line 246: I makes me wonder what are the performance implication of this solution. 
- When we are adding a single line only, then calling `shift` once could be the most performant solution :+1:
- When we are adding multiple lines, we will repeatedly call `shift` as many times as many new lines were read from the L2 node. Could this become a performance issue? 🤔  

We can use `slice` instead, but that's creating a new copy of the array, which may have its own performance issues.

```
❯ node
Welcome to Node.js v16.16.0.
Type ".help" for more information.
> [1,2,3,4,5].slice(-2)
[ 4, 5 ]
```

```js
  childLog = childLog.slice(-100)
```

I think the best solution is to use [`splice`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) and carefully calculate the offsets of items to delete.

```suggestion
  childLog.splice(0, childLog.length - 100)
```

WDYT?

I am debating with myself whether it's important or not to think about performance here. Your current solution is much easier to understand.

bajtos (approved):
It would be great to have an automated test to verify the fix.

I think writing such a test is not easy in the current way, as it requires running the L2 node. To make testing easier, we should extract the code keeping child process log lines into a new file, so that we can test the functionality in an isolated way (via unit-tests).

That's a bit more work than just a quick fix. Considering the severity of the issue you are fixing, I am ok to rely on manual testing for now.

main/saturn-node.js
bajtos: Example from REPL calling `splice` to keep only last 3 items:

```
❯ node
Welcome to Node.js v16.16.0.
Type ".help" for more information.
> x = [1, 2, 3, 4, 5]
[ 1, 2, 3, 4, 5 ]
> x.splice(0, x.length - 3)
[ 1, 2 ]
> x
[ 3, 4, 5 ]
```

juliangruber: I actually started with `slice`, but then switched to `shift` because of a micro benchmark:

```
> console.time('shift');{for(let i=0;i<1000;i++){const arr = new Array(105).fill();arr.shift();arr.shift();arr.shift();arr.shift();arr.shift()}}console.timeEnd('shift')
shift: 1.458ms
undefined
> console.time('slice');{for(let i=0;i<1000;i++){const arr = new Array(105).fill();arr.slice(-100)}}console.timeEnd('slice')
slice: 2.122ms
```

It's probably not a good benchmark (because of all the array allocations), but that was enough of a signal for me. Now that you brought up `splice` however, that is faster again:

```
> console.time('splice');{for(let i=0;i<1000;i++){const arr = new Array(105).fill();arr.splice(0, 5)}}console.timeEnd('splice')
splice: 0.943ms
```

I was wondering about performance here as well, it likely doesn't matter but since we both wondered let's just pick the faster implementation.

juliangruber:
> I think writing such a test is not easy in the current way, as it requires running the L2 node. To make testing easier, we should extract the code keeping child process log lines into a new file, so that we can test the functionality in an isolated way (via unit-tests).
>
> That's a bit more work than just a quick fix. Considering the severity of the issue you are fixing, I am ok to rely on manual testing for now.

Agreed 💯

bajtos (approved):
:shipit:

PR #245, opened 2022-10-25
Refactor dashboard initialization

jopedroliveira:
Improvement on the `Dashboard` page codebase as suggested on [#199 comment](https://github.com/filecoin-station/filecoin-station/pull/199#discussion_r989098138)
Parallelization of stored data requests

### DIFF
--- a/renderer/src/pages/Dashboard.tsx
+++ b/renderer/src/pages/Dashboard.tsx
@@ -32,10 +32,12 @@ const Dashboard = (): JSX.Element => {
 
   useEffect(() => {
     const loadStoredInfo = async () => {
-      Promise.all([getFilAddress().then(setAddress),
-        getAllActivities().then(setActivities),
-        getTotalEarnings().then(setTotalEarnigs),
-        getTotalJobsCompleted().then(setTotalJobs)])
+      Promise.all([
+        (async () => { setAddress(await getFilAddress()) })(),
+        (async () => { setActivities(await getAllActivities()) })(),
+        (async () => { setTotalEarnigs(await getTotalEarnings()) })(),
+        (async () => { setTotalJobs(await getTotalJobsCompleted()) })()
+      ])
     }
     loadStoredInfo()
 
--- a/renderer/src/pages/Dashboard.tsx
+++ b/renderer/src/pages/Dashboard.tsx
@@ -30,16 +30,15 @@ const Dashboard = (): JSX.Element => {
     navigate('/wallet', { replace: true })
   }
 
-  const reload = async (): Promise<void> => {
-    setAddress(await getFilAddress())
-    setActivities(await getAllActivities())
-    setTotalEarnigs(await getTotalEarnings())
-    setTotalJobs(await getTotalJobsCompleted())
-  }
-
-  useEffect(() => { reload() }, [])
-
   useEffect(() => {
+    const loadStoredInfo = async () => {
+      setAddress(await getFilAddress())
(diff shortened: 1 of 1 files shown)


bajtos (dismissed):

renderer/src/pages/Dashboard.tsx
```diff
@@ -30,16 +30,15 @@ const Dashboard = (): JSX.Element => {
...
-
-  useEffect(() => { reload() }, [])
-
   useEffect(() => {
+    const loadStoredInfo = async () => {
+      Promise.all([getFilAddress().then(setAddress),
+        getAllActivities().then(setActivities),
+        getTotalEarnings().then(setTotalEarnigs),
+        getTotalJobsCompleted().then(setTotalJobs)])
+    }
```
juliangruber, line 39: We're aiming to have async/await everywhere, this refactor is "downgrading" to Promise methods. Could you please revert that part?

juliangruber (changes requested):

renderer/src/pages/Dashboard.tsx
jopedroliveira: `Promisse.all` expects an `iterable` of promises. Using async/await we will be passing an iterable of method calls that won't run in parallel, but in series, which is the opposite of the desired here. 
This way, passing a `Promise`, it runs in parallel.
Thoughts?

jopedroliveira: I tried it, but the behavior is the same. It's still not a `Promise`.
I can measure the effect. 
Nevertheless, in order to implement the skeleton loading present on figma, this loading should be async. 
Let me know if this change makes no sense for you and I will revert it

juliangruber: > I tried it, but the behavior is the same. It's still not a `Promise`.

What do you mean by this?

This implementation should fetch all state in parallel:

```js
Promise.all([
  (async () => { setAddress(await getFilAddress()) })(),
  (async () => { setActivities(await getAllActivities()) })(),
  (async () => { setTotalEarnings(await getTotalEarnings()) })(),
  (async () => { setTotalJobs(await getTotalJobsCompleted()) })(),
])
```

bajtos: A possible alternative - just an idea to consider.

```ts
const [address, activities, totalEarnings, totalJobs] = await Promise.all([
  getFilAddress(),
  getAllActivities(),
  getTotalEarnings(),
  getTotalJobsCompleted()
])
setAddress(address)
//...
setTotalJobs(totalJobs)
```

jopedroliveira: > > I tried it, but the behavior is the same. It's still not a `Promise`.
> 
> What do you mean by this?
> 
> This implementation should update all state in parallel:


I did try a similar approach before, but I may have done something different. 
Your suggestion yields the expected parallelization as well ☺️ committed it on [713682e](https://github.com/filecoin-station/filecoin-station/pull/245/commits/713682e83ed1dd72adea1117d78d81637cae6f43)

juliangruber: > A possible alternative - just an idea to consider.

@bajtos I thought about this as well, and that would be the simplest way to go, however I assumed @jopedroliveira wanted to also parallelize the updating of state

juliangruber (approved):

juliangruber:
🥳

PR #253, opened 2022-10-31
Add report Sentry exit errors to Sentry. Closes #247

### LINKED ISSUE #247: Report Saturn L2 exit errors to Sentry
When the Saturn L2 Node exits with a non-zero code, we should report that code to Sentry an ideal include additional details, e.g. stderr content we have captured.

The activity log item should explain to the user that we have been notified about the problem. We should also tell them what to do to fix the problem or that no action is required on their side.

cleanshot_2022-10-26_at_00 43 30_2x

juliangruber:
Closes #247

### DIFF
--- a/main/saturn-node.js
+++ b/main/saturn-node.js
@@ -9,6 +9,7 @@ const { fetch } = require('undici')
 const fs = require('node:fs/promises')
 const path = require('path')
 const { setTimeout } = require('timers/promises')
+const Sentry = require('@sentry/node')
 
 const configStore = new Store()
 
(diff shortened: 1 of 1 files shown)


juliangruber:
Merging now and will publish a new release, in order to get better reporting

@bajtos hindsight please

main/saturn-node.js
```diff
@@ -159,6 +160,10 @@ async function start (/** @type {Context} */ ctx) {
     console.log(msg)
     appendToChildLog(msg)
     ctx.recordActivity({ source: 'Saturn', type: 'info', message: msg })
+    Sentry.captureException(new Error(msg), scope => {
+      scope.setExtra('logs', getLog())
```
bajtos, line 164: Maybe `saturnLogs` or `moduleLogs` would be a more descriptive key?

bajtos (commented):
LGTM 👍🏻

main/saturn-node.js
juliangruber: > Maybe `saturnLogs` or `moduleLogs` would be a more descriptive key?

Do you feel strongly about this? For as long as this is the only key we attach, I find it self explanatory enough.



> On second thought: When the process exits, the stdout and stderr streams may not be fully flushed yet. It would be more robust to report the logs from the `childProcess.on('close')` listener.

Great catch! https://github.com/filecoin-station/filecoin-station/pull/280

bajtos: > Do you feel strongly about this? For as long as this is the only key we attach, I find it self explanatory enough.

Fair enough 👍🏻

PR #254, opened 2022-10-31
Remove failing Saturn WebUI test

juliangruber:
This test is currently failing in #main. It's failing in CI, but not locally. Since we don't render Saturn's WebUI any more, I think we can just remove it without requiring further investigation.


juliangruber:
Merging this now to unblock other PRs

@bajtos hindsight please

bajtos (commented):
:+1:

PR #256, opened 2022-10-31
feat: skeleton loading

jopedroliveira:
This PR adds a _skeleton-loading_ component and its implementation for the Dashboard page, according to the [Figma prototype](https://www.figma.com/file/wQUmwtwIeak62uG60989XL/Station?node-id=1992%3A30314)

A preview can be found in the video below: 

https://user-images.githubusercontent.com/6967377/199051908-64b7a8b7-6e30-4c23-bbcb-ccb9e50d1ed2.mov

### DIFF
--- a/renderer/src/components/SkeletonLoading.tsx
+++ b/renderer/src/components/SkeletonLoading.tsx
@@ -1,0 +1,13 @@
+import { FC } from 'react'
+
+const SkeletonLoading: FC = () => {
+  return (
+    <div className='w-full h-8 rounded-md
+       bg-gradient-to-r from-[#ffffff00] via-[#ffffff57] to-[#ffffff00]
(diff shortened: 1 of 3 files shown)


juliangruber (changes requested):
I tried this PR, but am never seeing any animation, because the loading is too fast. Is this really required? Is this change effective on your side?

jopedroliveira:
> I tried this PR, but am never seeing any animation, because the loading is too fast. Is this really required? Is this change effective on your side?

It's not taking any effect, which is a good sign indeed. However, It should be a matter of time before we start experiencing some loading time, especially with activity growth. 
Other than that, this is part of the UI designs.

juliangruber:
We decided to keep this PR for reference, should we add skeleton loading later. For now, with our loading speeds, this isn't required 🙏

PR #274, opened 2022-11-18
Fix show UI when opening Station twice. Closes #266

### LINKED ISSUE #266: Cannot open GUI after restarting Computer
Started Filecoin station- amazing (can't wait to do this with Launchpad residents)

Tried just as an experiment to see what would happen if I restarted my comp. It seems as though Filecoin station has processes running, but I cannot open the GUI from the menu bar or Applications folder.

And also now somehow the icon is stuck on screen:
Screen Shot 2022-11-08 at 2 27 05 PM

What is the process/ application name so I can `killall` in terminal?

MacOS Monterey V12.6, Apple M1 Pro chip

juliangruber:
In #266, @walkerlj0 thought Station didn't open when she opened it, although it was already open (and hidden in the tray).

This change makes it so the app show its UI when you open it a second time, to try to recover from that situation.

Closes #266.

### DIFF
--- a/main/index.js
+++ b/main/index.js
@@ -111,6 +111,10 @@ app.on('before-quit', () => {
   Sentry.close()
 })
 
+app.on('second-instance', () => {
+  ctx.showUI()
+})
+
 process.on('uncaughtException', err => {
   Sentry.captureException(err)
   log.error(err)
--- a/main/index.js
+++ b/main/index.js
@@ -72,6 +72,10 @@ if (!app.requestSingleInstanceLock() && !inTest) {
   app.quit()
 }
 
+app.on('second-instance', () => {
+  ctx.showUI()
+})
+
 const jobStats = new JobStats()
 
 const activityLog = new ActivityLog()
@@ -109,10 +113,6 @@ app.on('before-quit', () => {
   // Flush pending events immediately
   // See https://docs.sentry.io/platforms/node/configuration/draining/
   Sentry.close()
-})
-
-app.on('second-instance', () => {
-  ctx.showUI()
 })
 
 process.on('uncaughtException', err => {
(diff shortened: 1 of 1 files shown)


juliangruber:
Converted to a draft because CI is failing

bajtos (changes requested):
Great UX improvement 👏🏻 

Let's make the implementation easier to reason about and more robust in edge cases 👇🏻

main/index.js
```diff
@@ -111,6 +111,10 @@ app.on('before-quit', () => {
   Sentry.close()
 })
 
+app.on('second-instance', () => {
+  ctx.showUI()
+})
+
```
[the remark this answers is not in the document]
juliangruber: These are really great suggestions! I implemented both :)

main/index.js
```diff
@@ -68,10 +68,14 @@ if (process.platform === 'win32') {
 }
 
 // Only one instance can run at a time
-if (!app.requestSingleInstanceLock() && !inTest) {
+if (!inTest && !app.requestSingleInstanceLock()) {
   app.quit()
 }
 
+app.on('second-instance', () => {
```
bajtos, line 75: _Nitpick: feel free to ignore._

I would add a comment to explain the UX flow we are implementing here. E.g.

```ts
// When the user attempts to start the app from the menu because 
// they don't notice the Station icon in the tray, then help 
// them finding the Station by showing our main window.
app.on('second-instance', () => {
```

After I wrote this comment, I realised it would be even nicer if the main window could also show a dialog explaining to the user that the Station has been running in the background and that they can always find it in the Tray (macOS menu bar). That's probably out of the scope of this PR, though.

bajtos (approved):
Much better 👍🏻

main/index.js
juliangruber: great suggestions, added docs and opened https://github.com/filecoin-station/filecoin-station/issues/282

PR #280, opened 2022-11-21
Fix submit Saturn logs once output streams flushed

juliangruber:
Ref https://github.com/filecoin-station/filecoin-station/pull/253#discussion_r1009862112


main/saturn-node.js
```diff
@@ -152,12 +152,15 @@ async function start (/** @type {Context} */ ctx) {
     console.log(msg)
     appendToChildLog(msg)
     ctx.recordActivity({ source: 'Saturn', type: 'info', message: msg })
-    Sentry.captureException(new Error(msg), scope => {
+
+    ready = false
+  })
+
+  childProcess.on('close', () => {
+    Sentry.captureException('Saturn node exited', scope => {
```
bajtos, line 160: Hmm, this is not going to tell us whether the Saturn node exited successfully (exit code 0 - I think we should not report this BTW), failed with an error (non-zero exit code) or crashed on something like SIGSEGV.

I think we should capture the exit reason in `childProcess.on('exit')` handler and then report that exit reason to Sentry.

```ts
// on('exit')
moduleExitReason = signal || code ? reason : null

// on('close')
if (moduleExitReason) {
  const msg = `Saturn node exited ${moduleExitReason}`
  Sentry.captureException(msg, scope => {
   // ...
  })
}
```

WDYT?

bajtos: Also inside `start()` function:

```ts
moduleExitReason = null // or undefined?
childLog = []
appendToChildLog('Starting Saturn node')
// etc.
```

juliangruber: > Hmm, this is not going to tell us whether the Saturn node exited successfully (exit code 0 - I think we should not report this BTW), failed with an error (non-zero exit code) or crashed on something like SIGSEGV.

Is a 0 exit code actually accepted? I believe this should never happen.

Otherwise I'm +1 to adding this change

bajtos: > Is a 0 exit code actually accepted? I believe this should never happen.

It happens when the node shuts down cleanly. In theory, this should be the case when the Station exits. In practice, I am not sure if the current implementation in Station is waiting for all modules to stop before exiting the app.

When emulate this scenario by running `ps ax | grep saturn` and then `kill <pid-of-saturn-L2-node>`, I get the following Activity Log entry:

Screenshot 2022-11-23 at 15 30 27

juliangruber: > When emulate this scenario by running `ps ax | grep saturn` and then `kill <pid-of-saturn-L2-node>`, I get the following Activity Log entry:

If saturn was `kill`ed, we'd definitely want to know, right? Otoh, it would be really nice if we could just rely on `0` being an ok exit code and us not having to do anything. Otherwise, if `0` can also mean trouble, we would always need to treat it as an error except if Station is exiting (remove `"close"` listener on before exit?). WDYT?

bajtos: > If saturn was `kill`ed, we'd definitely want to know, right? 

Hmm, you are right. We do want to know when saturn was `kill`ed.

> Otoh, it would be really nice if we could just rely on `0` being an ok exit code and us not having to do anything. Otherwise, if `0` can also mean trouble, we would always need to treat it as an error except if Station is exiting (remove `"close"` listener on before exit?). WDYT?

I think for now, let's always treat exit code `0` as a trouble and report it to Sentry.

For longer term, let's improve Station's shutdown sequence to wait for all modules to exit, ignore exit code 0 and report non-zero exit codes as errors. That would be a follow-up issue to be created.

For this PR, the remaining work is to save `moduleExitReason` in `exit` listener and include it in the error message reported to Sentry from `close` listener. So that we know whether the Saturn module exited cleanly or crashed.

WDYT?

juliangruber: sounds good all! like this?

bajtos (approved):
Great idea to capture the exit reason as an extra field.

I am not familiar with the way how Sentry visualises custom fields like this. Are they easy to find? I would prefer to include the exit reason in the message too. So that the cases when the Saturn module exits cleanly and when it exits with non-zero code are clean to spot in the list of errors and (hopefully?) create different Sentry error records too.

Anyhow, this is easy to tweak later once we have a better understanding of how Sentry deals with these error reports.

LGTM :shipit:

main/saturn-node.js
```diff
@@ -124,6 +126,7 @@ async function start (/** @type {Context} */ ctx) {
         console.log('Saturn node is up and ready (API URL: %s)', apiUrl)
         webUrl = `${apiUrl}webui`
         ready = true
+        moduleExitReason = null
```
bajtos, line 129: Actually, we need to clean `moduleExitReason` at the time when we start the process. If the process does not start successfully (never prints the `API:` line), then your implementation never cleans `moduleExitReason`.

On second thought, maybe we don't need to clear this flag since it should always be set from the `exit` event listener.

main/saturn-node.js
```diff
@@ -152,12 +155,17 @@ async function start (/** @type {Context} */ ctx) {
     console.log(msg)
     appendToChildLog(msg)
     ctx.recordActivity({ source: 'Saturn', type: 'info', message: msg })
-    Sentry.captureException(new Error(msg), scope => {
+
+    ready = false
+    moduleExitReason = signal || code ? reason : null
+  })
+
+  childProcess.on('close', () => {
```
bajtos, line 163: Do we need two `on('close')` listeners? Cannot we add this code to the existing listener on lines 142/145 above?

bajtos (commented):
Two more thoughts to consider.

main/saturn-node.js
juliangruber: Good call, I went with your suggestion to clear it more reliably

main/saturn-node.js
juliangruber: Good call, I merged the two 👍

juliangruber:
> I am not familiar with the way how Sentry visualises custom fields like this. Are they easy to find? I would prefer to include the exit reason in the message too. So that the cases when the Saturn module exits cleanly and when it exits with non-zero code are clean to spot in the list of errors and (hopefully?) create different Sentry error records too.
> 
> Anyhow, this is easy to tweak later once we have a better understanding of how Sentry deals with these error reports.

Let's indeed tweak it as we go. To me storing this as a field is more useful because you can filter by field values for example, instead of having to manually parse the error description. You're right however that it could be more useful to include in the error message itself. Maybe we can also do both. Let's tweak as we go

juliangruber:
Let's retry CI when the L1 testnet has healthy nodes again

bajtos:
> Let's indeed tweak it as we go. To me storing this as a field is more useful because you can filter by field values for example, instead of having to manually parse the error description. You're right however that it could be more useful to include in the error message itself. Maybe we can also do both. Let's tweak as we go

I was thinking about including the exit code in both places - in the error message and as a custom field.

Let's wait until we have some error reports in Sentry and then decide if we need to make any adjustments to your current implementation.

Issue #296, opened 2022-12-01
Update all dependencies to their latest version


bajtos:
Somehow, Dependabot is not sending pull request for all of our dependencies.

Let's update them manually.

Optionally, investigate why Dependabot is not opening PRs for these dependencies.

```
❯ npm outdated --depth 0
Package                                Current  Wanted  Latest  Location                                            Depended by
@glif/filecoin-address                   2.0.1  2.0.20  2.0.20  node_modules/@glif/filecoin-address                 filecoin-station
@playwright/test                        1.27.1  1.28.1  1.28.1  node_modules/@playwright/test                       filecoin-station
@sentry/node                            7.21.1  7.22.0  7.22.0  node_modules/@sentry/node                           filecoin-station
@sentry/react                           7.21.1  7.22.0  7.22.0  node_modules/@sentry/react                          filecoin-station
@sentry/tracing                         7.21.1  7.22.0  7.22.0  node_modules/@sentry/tracing                        filecoin-station
@types/mocha                            10.0.0  10.0.1  10.0.1  node_modules/@types/mocha                           filecoin-station
@types/react-dom                        18.0.8  18.0.9  18.0.9  node_modules/@types/react-dom                       filecoin-station
@typescript-eslint/eslint-plugin        5.39.0  5.45.0  5.45.0  node_modules/@typescript-eslint/eslint-plugin       filecoin-station
@typescript-eslint/experimental-utils   5.42.1  5.45.0  5.45.0  node_modules/@typescript-eslint/experimental-utils  filecoin-station
@typescript-eslint/parser               5.42.1  5.45.0  5.45.0  node_modules/@typescript-eslint/parser              filecoin-station
@vitejs/plugin-react                     2.1.0   2.2.0   2.2.0  node_modules/@vitejs/plugin-react                   filecoin-station
@vitest/ui                              0.24.3  0.24.5  0.25.3  node_modules/@vitest/ui                             filecoin-station
dayjs                                   1.11.5  1.11.6  1.11.6  node_modules/dayjs                                  filecoin-station
electron                                21.3.1  21.3.3  22.0.0  node_modules/electron                               filecoin-station
execa                                    5.1.1   5.1.1   6.1.0  node_modules/execa                                  filecoin-station
jsdom                                   20.0.2  20.0.3  20.0.3  node_modules/jsdom                                  filecoin-station
playwright                              1.27.1  1.28.1  1.28.1  node_modules/playwright                             filecoin-station
[...]

PR #302, opened 2022-12-05
build: run dependabot every weekday

bajtos:
See the discussion in #296

> We have configured Dependabot to run once a week. That is great for avoiding too much noise caused by dependency updates. The downside is that we can update only five dependencies every week, which is clearly not enough to keep up with the pace.

In this PR, I am changing dependabot's config to run daily (Monday-Friday) at 9:00 am CET.


juliangruber (approved):
nice!

PR #310, opened 2022-12-06
fix: always build for x64 on macOS

bajtos:
Before this change, when running `npm run package` on an Apple M1/M2 computer, electron builder was creating a package for `arm64`. Since we are not downloading Saturn L2 node for `arm64`, we ended up with a broken Station app.

In this commit, I am changing electron-builder config to always build for `x64` when running on macOS.


bajtos:
Extracted from https://github.com/filecoin-station/filecoin-station/pull/294

juliangruber (approved):

PR #332, opened 2022-12-12
refactor: move station activity jobs rpc into hook

jopedroliveira:
This PR moves activity-related API calls into a dedicated hook. Also splits RPC calls into multiple `useEffect` instead of anonymous functions being called inside `Promise.all`


joaojusto (approved):
LGTM

juliangruber (approved):

PR #342, opened 2022-12-14
feat: improve logging

bajtos:
- feat: log auto-updater messages
- feat: log station-config messages to log file
- feat: log Station build version and computer info

Example stdout output:

```
Log file: /Users/bajtos/Library/Logs/Filecoin Station/main.log
11:48:33.975 (config) › Loading Station configuration from /Users/bajtos/Library/Application Support/Filecoin Station/config.json
11:48:34.004 (main)    › Filecoin Station build version: 0.11.3.1-dev [DEV]
11:48:34.004 (main)    › Running on darwin arm64 version 22.1.0
(node:17388) DeprecationWarning: findLogPath() is deprecated and will be removed in v5.
```

Notice the deprecation warning about `findLogPath()`. The documentation does not explain what we should use instead of `findLogPath()` 🙁

With my changes in place, we can improve our UI for reporting problems and allow users to send their electron-log file, too. Perhaps in addition to the Saturn log file. Or we can change the Saturn Module to log messages using electron-log, so that we have everything in one place. Either way, that's out of scope of my PR.

### DIFF
--- a/main/index.js
+++ b/main/index.js
@@ -1,8 +1,11 @@
 'use strict'
 
 const { app, dialog } = require('electron')
-const log = require('electron-log')
+const electronLog = require('electron-log')
 const path = require('node:path')
+
+console.log('Log file:', electronLog.transports.file.findLogPath())
+const log = electronLog.scope('main')
 
 // Override the place where we look for config files when running the end-to-end test suite.
 // We must call this early on, before any of our modules accesses the config store.
@@ -23,6 +26,7 @@ const { ActivityLog } = require('./activity-log')
 const { BUILD_VERSION } = require('./consts')
 const { JobStats } = require('./job-stats')
 const { ipcMain } = require('electron/main')
+const os = require('os')
 const saturnNode = require('./saturn-node')
 const serve = require('electron-serve')
 const { setupAppMenu } = require('./app-menu')
@@ -38,7 +42,8 @@ const { setup: setupDialogs } = require('./dialog')
 const inTest = (process.env.NODE_ENV === 'test')
 const isDev = !app.isPackaged && !inTest
 
-console.log('Filecoin Station build version:', BUILD_VERSION)
+log.info('Filecoin Station build version:', BUILD_VERSION, isDev ? '[DEV]' : '', inTest ? '[TEST]' : '')
+log.info('Running on %s %s version %s', os.platform(), os.arch(), os.release())
 
 // Expose additional metadata for Electron preload script
 process.env.STATION_BUILD_VERSION = BUILD_VERSION
--- a/main/station-config.js
+++ b/main/station-config.js
@@ -3,6 +3,8 @@
 const Store = require('electron-store')
 const { randomUUID } = require('crypto')
 
+const log = require('electron-log').scope('config')
+
 const ConfigKeys = {
   OnboardingCompleted: 'station.OnboardingCompleted',
   TrayOperationExplained: 'station.TrayOperationExplained',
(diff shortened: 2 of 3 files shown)


main/index.js
```diff
@@ -1,9 +1,12 @@
 'use strict'
 
 const { app, dialog } = require('electron')
-const log = require('electron-log')
+const electronLog = require('electron-log')
 const path = require('node:path')
 
+console.log('Log file:', electronLog.transports.file.findLogPath())
```
bajtos, line 7: My intention here is to make it easier to locate the electron-log file when troubleshooting problems. I am using `console.log` instead of electron-log because it feels redundant to log the filename of the log file into that log file.

Let me know if you thing otherwise!

main/index.js
```diff
@@ -38,7 +42,8 @@ const { setup: setupDialogs } = require('./dialog')
 const inTest = (process.env.NODE_ENV === 'test')
 const isDev = !app.isPackaged && !inTest
 
-console.log('Filecoin Station build version:', BUILD_VERSION)
+log.info('Filecoin Station build version:', BUILD_VERSION, isDev ? '[DEV]' : '', inTest ? '[TEST]' : '')
+log.info('Running on %s %s version %s', os.platform(), os.arch(), os.release())
```
bajtos, line 46: I am not sure which APIs are better to use. Do you have any opinion on this?

- `os.arch()` vs `os.machine()`
- `os.platform()` vs `os.type()`

The former APIs describe what is the target for which Node.js was built. The latter APIs make sys calls to determine the value.

Example output:

```
> console.log({arch: os.arch(), machine: os.machine(), platform: os.platform(), type: os.type()})
{ arch: 'arm64', machine: 'arm64', platform: 'darwin', type: 'Darwin' }
```

Node.js API docs:

https://nodejs.org/dist/latest-v18.x/docs/api/os.html#osarch
> Returns the operating system CPU architecture for which the Node.js binary was compiled. Possible values are 'arm', 'arm64', 'ia32', 'mips', 'mipsel', 'ppc', 'ppc64', 's390', 's390x', and 'x64'.

https://nodejs.org/dist/latest-v18.x/docs/api/os.html#osmachine
> Returns the machine type as a string, such as arm, aarch64, mips, mips64, ppc64, ppc64le, s390, s390x, i386, i686, x86_64.

https://nodejs.org/dist/latest-v18.x/docs/api/os.html#osplatform
> Returns a string identifying the operating system platform for which the Node.js binary was compiled. The value is set at compile time. Possible values are 'aix', 'darwin', 'freebsd','linux', 'openbsd', 'sunos', and 'win32'.

https://nodejs.org/dist/latest-v18.x/docs/api/os.html#ostype
> Returns the operating system name as returned by [uname(3)](https://linux.die.net/man/3/uname). For example, it returns 'Linux' on Linux, 'Darwin' on macOS, and 'Windows_NT' on Windows.

bajtos: Let's print both. The build target can be different from the computer architecture, e.g. when running Darwin x86_64 binaries on arm64 via Rosetta.

bajtos: Unfortunately, `os.machine()` has been added in Node.js v18, which is not available in Electron yet. Hopefully, we will get this in Feb 2023 - see https://github.com/electron/electron/pull/36356 and https://www.electronjs.org/docs/latest/tutorial/electron-timelines

Screenshot 2022-12-14 at 15 03 27

juliangruber (approved):
Great work!

PR #361, opened 2023-01-02
Fix Sentry UI log overflow

juliangruber:
Screenshot 2023-01-02 at 17 49 15

https://sentry.io/organizations/protocol-labs-ip/issues/3803942762/?project=6762462&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=7d

I kind of arbitrarily went with 10 lines, I don't know what their actual limit is. The UI seems to cap after 17446 characters. Another way to fix this would be to reverse the sorting, so that the latest log lines appear first.


bajtos:
Looks good at the high level 👍🏻 

It makes me wonder, though, whether the complexity of adding `maxLines` argument to the `getLog` function is worth it - maybe it would be simpler to call `childLog.slice().join()`?

More importantly, `getLog` is used also for IPC, you need to fix the following line flagged by TypeScript compiler:

```ts
 ipcMain.handle('saturn:getLog', saturnNode.getLog)
```

bajtos (commented):
☝🏻

bajtos (approved):
:shipit:

PR #373, opened 2023-01-04
Add Wallet Backend. #388

juliangruber:
For #388.


@types/@zondax/filecoin-signing-tools/index.d.ts
```diff
@@ -0,0 +1,241 @@
+declare module '@zondax/filecoin-signing-tools' {
```
bajtos, line 1: Where are these `.d.ts` files coming from? Did you write them yourself?

It would be great to eventually contribute these typings to `@zondax/filecoin-signing-tools` to make them useful for the wider community and so that we don't have to maintain them on our own.

Maybe we can open a follow-up issue in our repo for this task?

package.json
```diff
@@ -39,17 +39,20 @@
   },
   "homepage": "https://github.com/filecoin-project/filecoin-station#readme",
   "dependencies": {
-    "@glif/filecoin-address": "^2.0.1",
+    "@glif/filecoin-address": "^2.0.20",
```
bajtos, line 42: Is it necessary to bump `@glif/filecoin-address` from `2.0.1` to `2.0.20`? I think this is the root cause of the Buffer problem, isn't it? Maybe we should solve this upgrade first in https://github.com/filecoin-station/filecoin-station/pull/240, in isolation from other changes in this PR. What do you think?

bajtos: Downgrading this module back to `2.0.1` seems to fix the problem. The GUI starts fine, and `npm run build` works too.

I noticed one warning in the DevTools console, I think it's not a problem, but it could be a hint on how to fix the Buffer issue for the new Filecoin-address version.

Screenshot 2023-01-05 at 14 37 43

bajtos: See https://github.com/filecoin-station/filecoin-station/pull/379

main/saturn-node.js
```diff
@@ -70,8 +70,8 @@ function getSaturnBinaryPath () {
 }
 
 async function start (/** @type {Context} */ ctx) {
-  if (!getFilAddress()) {
-    console.info('Saturn node requires FIL address. Please configure it in the Station UI.')
+  if (!wallet.getAddress()) {
+    console.info('Saturn node requires FIL address')
```
bajtos, line 74: This should never happen, right?

I am proposing to change `console.info` to `console.error`, or even better use electron-log with error severity to capture this information in the log file for future inspection.

It makes me wonder if we should throw an exception or exit the station completely if we ever get here.

main/station-config.js
```diff
@@ -73,35 +72,13 @@ function setTrayOperationExplained () {
   configStore.set(ConfigKeys.TrayOperationExplained, TrayOperationExplained)
 }
 
-/**
- * @returns {string | undefined}
- */
-function getFilAddress () {
-  return FilAddress || 'f0111' // needs refactor - same as getStationWalletAddress
```
bajtos, line 80: Is it worth our effort to implement a migration that will read the FIL address provided by the user and set it as the destination address in the wallet?

main/wallet.js
```diff
@@ -0,0 +1,52 @@
...
+    1,
+    CoinType.MAIN
+  )
+  log.info('Address: %s', address)
+}
+
+/**
+ * @returns {string}
+ */
+function getAddress () {
```
bajtos, line 45: When reading code calling `wallet.getAddress()`, I am not sure whether it's the address of the local wallet or the destination address configured in the Wallet UI.

I think a name like `getStationAddress()` or `getLocalAddress` may work better.

bajtos (approved):
LGTM 👍🏻

renderer/src/components/WalletModule.tsx
```diff
@@ -64,7 +64,7 @@ const WalletModule: FC<PropsWallet> = ({ isOpen = false }) => {
     <div className='relative z-20'>
       <div className='h-8 bg-primary-dark flex items-center px-8' onClick={reset}>
         <p className='text text-body-3xs text-white opacity-80 mr-3'>STATION ADDRESS</p>
-        <p className='text text-body-3xs text-white'>{stationAddress}</p>
+        <p className='text text-body-3xs text-white'>{`${stationAddress.slice(0, 6)} ... ${stationAddress.slice(-6)}`}</p>
```
bajtos, line 67: Let's use the ellipsis character ('…', U+2026 HORIZONTAL ELLIPSIS) instead of `...`.

See e.g. https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow and https://www.htmlsymbol.com/html-entities/horizontal-ellipsis.html

```suggestion
        <p className='text text-body-3xs text-white'>{`${stationAddress.slice(0, 6)} &hellip; ${stationAddress.slice(-6)}`}</p>
```

@types/@zondax/filecoin-signing-tools/index.d.ts
juliangruber: I copied these over from https://github.com/glifio/react-components/tree/primary/%40types/%40zondax, that repo also being the reason we're using this module in the first place.

Good call, I opened https://github.com/filecoin-station/filecoin-station/issues/402

package.json
juliangruber: the downgrade worked 🙌 

Also the `bufer.Buffer` warning seems not to create any issues 👍

main/saturn-node.js
juliangruber: Based on the timing of calls in `main/index.js`, this can never happen indeed. The condition is here to satisfy the type system.

I think you're right that Station should crash hard here. I've changed it to a `throw`.

main/station-config.js
juliangruber: I actually don't think so. It could be confusing for the user even. If we don't add it, everyone will go through the same flow to set up their wallet.

main/wallet.js
juliangruber: Hm, I'm surprised by this, since to me `wallet` by definition is the only wallet available to Station. And that is the internal one. And a wallet only has one address, so `wallet.getStationAddress()` doesn't make so much sense to me. If anything I think we could rename it to `stationWallet.getAddress()`, however that's prefixing our own components with our own name, which is kind of redundant. What do you think?

renderer/src/components/WalletModule.tsx
juliangruber: great idea, done!

main/wallet.js
bajtos: I see. Your explanation makes sense 👍🏻 Let's keep things as they are now.

main/typings.d.ts
```diff
@@ -10,13 +10,24 @@ export interface Activity {
   message: string;
 }
 
+export type FILTransactionProcessing = {
+  hash: string;
+  height: number;
+  timestamp: number?;
+  status: TransactionStatus?;
```
bajtos, line 17:
```suggestion
  timestamp?: number;
  status?: TransactionStatus;
```

main/typings.d.ts
```diff
@@ -10,13 +10,24 @@ export interface Activity {
...
+  outgoing: boolean;
+  amount: string;
+  address: string;
+}
+
 export type FILTransaction = {
```
bajtos, line 23: Would something like this be a meaningful way how to avoid duplication?

```ts
export type FILTransaction = FILTransactionProcessing &
  Required<Pick<FILTransactionProcessing, 'timestamp' | 'status'>>
```

Explanation:
- The generic type `Required` converts all fields from optional (e.g. `number | undefined`) to required (`number`).

- The generic type `Pick` picks a subset of properties from the target object-like type.

- Together, they create a type with two required properties - `timestamp` and `status`. 

Personally, I would design these types the other way round - start with `FILTransaction` with all fields required, and then define `FILTransactionProcessing` by making some fields optional.

```ts
// helper
type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>

// FILTransaction with certain properties changed to optional
type FILTransactionProcessing = PartialBy<FILTransaction, 'timestamp' | 'status'>
```

BTW, should it be `FILTransactionProcessing` or `FILTransactionPending`?

juliangruber: Great suggestion! I'll refactor it like this 🙏

juliangruber: the first suggestion works, with the 2nd I get this error:

Screenshot 2023-01-24 at 16 03 44

Implementation:

```ts
export type FILTransaction = {
  hash: string;
  height: number;
  timestamp: number;
  status: TransactionStatus;
  outgoing: boolean;
  amount: string;
  address: string;
}

// helper
type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>

// FILTransaction with certain properties changed to optional
export type FILTransactionProcessing = PartialBy<FILTransaction, 'hash' | 'timestamp' | 'status'>
```

Do you know how to get that to work? I would also prefer the 2nd way

juliangruber: nvm, just needed to remove the null values 👍 

Thank you so much!

main/wallet.js
```diff
@@ -0,0 +1,435 @@
...
+    ) {
+      const { hash } = transaction
+      stateReplaysBeingFetched.add(hash)
+      transaction.status = 'processing'
+      ;(async () => {
+        while (true) {
+          try {
+            const stateReplay = await getStateReplay(hash)
+            transaction.status = stateReplay.receipt.exitCode === 0
+              ? 'sent'
```
bajtos, line 233: I think the status `sent` can be confusing when it's shown for incoming transactions (even though they are unlikely).

Can we find a better name for this status? How about `confirmed` or `committed`?

renderer/src/hooks/StationWallet.tsx
```diff
@@ -59,38 +89,58 @@ const useWallet = (): Wallet => {
...
+}
+
+interface SplitTransactions {
+  processing: FILTransactionProcessing | undefined,
+  confirmed: FILTransaction[]
+}
+
+const splitWalletTransactions = (transactions: (FILTransaction|FILTransactionProcessing)[]): SplitTransactions => {
+  const processing: FILTransactionProcessing | null =
+  transactions.filter(tx => tx.status === 'processing')[0] as FILTransactionProcessing
```
bajtos, line 140: I would consider implementing a [Type Guard function](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) to discriminate between different transaction types.

```ts
//
// inside typings
//

export function isFILTransactionProcessing(tx: FILTransaction | FILTransactionProcessing): tx is FILTransactionProcessing {
  return tx.status === 'processing'
}

export function isFILTransactionConfirmed(tx: FILTransaction | FILTransactionProcessing): tx is FILTransaction {
  return tx.status !== 'processing'
}

//
// usage here
//

const processing = transactions.filter(isFILTransactionProcessing)[0]
// btw, is it better to use transactions.find() instead?
const processing = transactions.find(isFILTransactionProcessing)

const confirmed = transactions.find(isFILTransactionConfirmed)
```

main/typings.d.ts
```diff
@@ -11,12 +11,52 @@ export interface Activity {
...
+  outgoing: boolean;
+  amount: string;
+  address: string;
+}
+
+// helper
+type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
+
+// FILTransaction with certain properties changed to optional
+export type FILTransactionProcessing = PartialBy<FILTransaction, 'hash' | 'height'>
```
bajtos, line 27: IIUC, we use this type when the transaction has `status` set to `processing`. You may want to encode that fact in the type too.

```suggestion
export type FILTransactionProcessing = PartialBy<FILTransaction, 'hash' | 'height'> & { status: 'processing' };
```

renderer/src/typings.d.ts
```diff
@@ -62,10 +62,18 @@ export type ActivityEventMessage = {
...
+  amount: string;
+  address: string;
 }
+
+// helper
+type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
+
+// FILTransaction with certain properties changed to optional
+type FILTransactionProcessing = PartialBy<FILTransaction, 'hash' | 'height'>
+type FILTransactionLoading = PartialBy<FILTransaction, 'status' | 'timestamp'>
```
bajtos, line 79: Same comments as for the typings in `main`.

Is it perhaps time to create a shared typings file that can be used by both `main` and `renderer`?

renderer/src/typings.d.ts
```diff
@@ -62,10 +62,18 @@ export type ActivityEventMessage = {
...
-  outgoing: boolean
-  amount: string
-  address: string
+  hash: string;
+  height: number;
+  timestamp: number;
+  status: TransactionStatus;
+  outgoing: boolean;
+  amount: string;
+  address: string;
```
bajtos, line 71: Semicolons or not? (See the discussion in `main` typings.)

bajtos (changes requested):
The implementation looks great overall!

I don't have enough knowledge about the wallet APIs to review all low-level details. I don't see any obvious problems though.

**There is one big missing piece though: automated tests.**

We need to ensure our code works as intended, the external Wallet API provides the same interface we are relying on, and future pull requests don't accidentally introduce regressions.

We can keep the tests lightweight for now, but it's crucial to have the testing infrastructure in place from the beginning, to make it easy to add new tests in future pull requests, and to verify that our code is designed and structured in a way that makes testing feasible.

I am proposing the following test cases to cover as the minimum set:
- Create a new wallet by generating a new seed, this should include the Filecoin Glif client.
- Nice to have: create a wallet by retrieving the seed phrase from keytar.
- `getBalance()` + `updateBalance()`
- `listTransactions()` + `updateTransactions()`
- Updating the transactions second time, when we want to fetch only new updates
- I think we should test `transferFunds` too. Is there a reasonably-easy way how to configure our wallet to use testnet and automatically get some funds from a faucet?

In order to be able to test these features in isolation from the rest of the Electron app, I think you may want to consider extracting the code dealing with Glif API into a standalone wallet backend - either a new file or maybe a class exported from the current file. Then the remaining code in `main/wallet.js` should be a relatively small glue between our wallet backend and the electron bits.

Something along the following lines. Of course, you will need to tweak the API depending on what you need for the ergonomy of testing and the ergonomy of the glue code.

```ts
// main/wallet-backend.js
export class StationWallet {
  constructor(opts) {
    // load the seed phrase, create the client
    // opts is an object with config overrides useful
    // for testing, e.g. 
    //  - testnet vs mainnet
    /   - hard-coded address to use
  }

  getAddress() { /* ... */ }

  async fetchBalance() { /* ... */ }
  async fetchAllTransactions() { /* ... */ } 
  async transferFunds(from, to, amount) { /* ... */ }
}
```

main/wallet.js
```diff
@@ -0,0 +1,435 @@
...
+
+    if (
+      transaction.hash &&
+      (!transaction.status || transaction.status === 'processing') &&
+      !stateReplaysBeingFetched.has(transaction.hash)
+    ) {
+      const { hash } = transaction
+      stateReplaysBeingFetched.add(hash)
+      transaction.status = 'processing'
+      ;(async () => {
```
bajtos, line 228: Here we are starting a background task to get the tipset and the state replay, creating one task for every transaction we need to update, and making two API requests for each one.

I am concerned about the performance impact if there are more than a handful of such transactions.

The code is also mixing the implementation of the functionality we need (fetch data, transform it to the shape we need, update our internal transaction representation) with a logic to retry the requests on errors.

IMO, this design makes this code difficult to reason about and test.

For example, now that I am reading through the code more carefully again, I spotted that you are calling `updateBalance` every time you encounter a transaction that changed the state to `sent`. It's ok in the "average" case when zero or one transaction changes state, but it maybe become problematic when updating more than one transaction.

I would prefer to design the "update" functionality as follows:
- Make error statuses (e.g. we could not fetch transaction status) explicit. Store them inside the transaction instead of the current design, where the state of our code implicitly remembers them.
- Change the "update" function to be a one-pass that finishes quickly.
  1. Collect list of all transactions
  2. Find transactions we need to update
  3. Find the list of tipsets we need to fetch.
  4. Find the list of state replays we need to fetch.
  5. Fetch tipsets with controlled concurrency. 
  6. Fetch state replays with controlled concurrency. 
  7. Maybe we can do 5 & 6 serially for now (`concurrency=1`)? If the API call fails, then simply skip updating the transaction, or maybe mark the transaction status to be "could not fetch update"
  8. Update transactions using the data from 6 & 7.
  9. If at least one transaction changed the state to 'sent', then call `updateBalance` (only once).
 
- It's up to the wallet glue code to decide how often to call the "update" function - maybe more often to retry failed API calls and less often when all went well.

BTW is it safe to call `updateBalance` multiple times at the same time? I think it creates a race condition:
1. Caller A calls `updateBalance`
2. Caller B calls `updateBalance`
3. It happens that API requests finish in a different order.
4. Caller B updates the balance using the value returned by the later API call.
5. Caller A updates the balance using the value returned by the earlier API call.
6. If a transaction was committed between A & B calls, our balance shows an outdated amount.

main/typings.d.ts
```diff
@@ -11,12 +11,52 @@ export interface Activity {
...
-  amount: string
-  address: string
+  hash: string;
+  height: number;
+  timestamp: number;
+  status: TransactionStatus;
+  outgoing: boolean;
+  amount: string;
+  address: string;
+}
```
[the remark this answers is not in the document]
juliangruber: This is a tricky one: The imported types don't use semicolons, `main/typings.d.ts` uses semicolons, `renderer/typings.d.ts` doesn't use semicolons. I've updated the style to be at least consistent per file. Since our linter didn't complain, I think this discussion is out of scope for this PR, I'm going to open a new issue 👍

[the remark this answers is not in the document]
juliangruber: https://github.com/filecoin-station/filecoin-station/issues/444

renderer/src/hooks/StationWallet.tsx
juliangruber: all good points! I've implemented it all :)

main/wallet.js
juliangruber: This is a good point! I changed it to `succeeded`, because I think a failed transaction can also be confirmed

main/typings.d.ts
juliangruber: Screenshot 2023-01-30 at 15 10 32

If a transaction starts as processing and then is market as failed, how can I convert the type? Or shall I recreate the transaction object?

renderer/src/typings.d.ts
juliangruber: I agree! I've created an issue for this: https://github.com/filecoin-station/filecoin-station/issues/445

main/wallet.js
```diff
@@ -0,0 +1,435 @@
...
+    apiAddress: 'https://api.node.glif.io/rpc/v0'
+  })
+  ;[address] = await provider.wallet.getAccounts(
+    0,
+    1,
+    CoinType.MAIN
+  )
+  log.info('Address: %s', address)
+
+  ;(async () => {
```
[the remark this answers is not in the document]
juliangruber: I've refactored this to add the `updateWalletData` function, I think that's an improvement, but I don't see it the same way about `setInterval`. Now that we have `async/await` available, `setInterval` isn't necessary any more and can be forgotten when writing any new code. `setInterval` is a platform feature, while you can do the same thing just with language features too. I think `setInterval` was invented because we didn't have `async/await` yet, but now it can be decomissioned.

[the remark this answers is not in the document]
juliangruber: > BTW should we fetch the balance & the list of transactions concurrently?

I think it's better UX not to do it, so that the balance always changes before the transactions, otherwise you might think a transaction didn't have any effect. Ideally we always show a consistent state, but that's not possible with these APIs

renderer/src/hooks/StationWallet.tsx
```diff
@@ -138,10 +143,10 @@ interface SplitTransactions {
 }
 
 const splitWalletTransactions = (transactions: (FILTransaction|FILTransactionProcessing)[]): SplitTransactions => {
-  const processing: FILTransactionProcessing | null =
-  transactions.find(tx => tx.status === 'processing') as FILTransactionProcessing
+  const processing: FILTransactionProcessing | undefined =
+  transactions.find(isFILTransactionProcessing)
```
bajtos, line 147: I think you don't need the explicit type declaration now.

```suggestion
  const processing = transactions.find(isFILTransactionProcessing)
```

Of course, feel free to keep it if you find it helpful to have the type of variable explicitly declared in the spruce code.

renderer/src/hooks/StationWallet.tsx
```diff
@@ -138,10 +143,10 @@ interface SplitTransactions {
...
   const confirmed: FILTransaction[] =
-  transactions.filter(tx => tx.status !== 'processing') as FILTransaction[]
+  transactions.filter(isFILTransactionConfirmed)
```
bajtos, line 149: Ditto. I think the following should work too.

```ts
const confirmed = transactions.filter(isFILTransactionConfirmed)
```

main/wallet.js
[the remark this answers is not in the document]
bajtos: > I think it's better UX not to do it, so that the balance always changes before the transactions

Makes sense 👍🏻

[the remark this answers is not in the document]
bajtos: TL;DR: **I am fine to keep your current version. Extracting `refreshState` helped a lot 👍🏻**

> I've refactored this to add the `updateWalletData` function, I think that's an improvement, but I don't see it the same way about `setInterval`. Now that we have `async/await` available, `setInterval` isn't necessary any more and can be forgotten when writing any new code. `setInterval` is a platform feature, while you can do the same thing just with language features too. I think `setInterval` was invented because we didn't have `async/await` yet, but now it can be decommissioned.

Well, you cannot avoid platform APIs entirely-  you still need `timers.setTimeout`.

```js
// I like this more:
setInterval(doSomeWork, 10_000)

// Rather than
;(async () => {
  while (true) { 
    await doSomeWork()
    await timers.setTimeout(10_000)
  }
})()
```

Maybe you are missing a helper function that would allow us to write something like this?

```js
repeatedlyRunInBackground(refreshState, 10_000)

// implementation
function repeatedlyRunInBackground(taskFn, intervalInMs) {
  ;(async () => {
    while (true) { 
      await taskFn()
      await timers.setTimeout(intervalInMs)
    }
  })()
}
```

Anyhow. This is a relatively minor detail. The code is isolated and easy to rework later if we ever change our mind.

I am fine to keep your current version. Extracting `refreshState` helped a lot 👍🏻

bajtos (commented):
Good progress 👏🏻

renderer/src/hooks/StationWallet.tsx
bajtos: (Assuming you like this terse version more than the original.)

juliangruber: I do :)

juliangruber:
The semantics are a bit different, because setInterval doesn't wait for `doSomeWork` to finish, and therefore can run multiple invocations at the same time. The 1:1 translation would be:

```js
;(async () => {
  while (true) { 
    doSomeWork().catch(...)
    await timers.setTimeout(10_000)
  }
})()
```

To me this example shows how much more flexible this more native JS approach is

juliangruber:
> **There is one big missing piece though: automated tests.**

Great point! I've followed your structure and ported as much as I could for now. Could you please take another look? The functions that I didn't port were too intertwined with electron/ipc/state logic. If you see ways to port more, could you please make specific suggestions?

We can't really test transferring funds, because
- it takes too long (> 20 minutes)
- the glif apis for testnet didn't work the same way as the mainnet ones

juliangruber:
@bajtos I had to revert the type guard PR, because while it worked during `npm run lint`, it failed with `npm start`:

```
6:09:34 PM [vite] Internal server error: Failed to resolve import "../typings" from "renderer/src/hooks/StationWallet.tsx". Does the file exist?
  Plugin: vite:import-analysis
  File: /Users/julian/dev/filecoin-station/filecoin-station/renderer/src/hooks/StationWallet.tsx:15:7
  2  |  import { useState, useEffect, useCallback } from "react";
  3  |  import { getDestinationWalletAddress, setDestinationWalletAddress, getStationWalletAddress, getStationWalletBalance, getStationWalletTransactionsHistory, transferAllFundsToDestinationWallet } from "../lib/station-config";
  4  |  import { isFILTransactionConfirmed, isFILTransactionProcessing } from "../typings";
     |                                                                         ^
  5  |  const useWallet = () => {
  6  |    _s();
```

Do you know what could be going on there? Also weird that in line 4 you only see the function imports, not the type imports 🤔

main/wallet.js
juliangruber: Updated like this <3 I also debounced `updateBalance`

> If at least one transaction changed the state to 'sent', then call updateBalance (only once).

We still need to update the balance periodically, as it can update way before a transaction gets confirmed (don't ask me how)

main/wallet.js
```diff
@@ -0,0 +1,205 @@
...
+
+/*
+ * @returns {Promise<void>}
+ */
+async function transferAllFundsToDestinationWallet () {
+  assert(backend.address)
+  const to = getDestinationWalletAddress()
+  assert(to)
+  // FIXME: Only transfer a little FIL for now
+  const balance = new FilecoinNumber('0.00001', 'fil')
```
juliangruber, line 167: Fix before landing

PR #406, opened 2023-01-16
fix unactionable Sentry errors

juliangruber:
It's not really actionable for us when Influx is slow or unreachable (can be either Influx or the client's connection)

### DIFF
--- a/main/telemetry.js
+++ b/main/telemetry.js
@@ -22,7 +22,9 @@ const writeClient = client.getWriteApi(
 
 setInterval(() => {
   writeClient.flush().catch(err => {
-    Sentry.captureException(err)
+    if (!/HttpError|getAddrInfo|RequestTimedOutError/i.test(String(err))) {
+      Sentry.captureException(err)
+    }
   })
 }, 5000).unref()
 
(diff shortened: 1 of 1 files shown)


main/telemetry.js
```diff
@@ -22,7 +22,9 @@ const writeClient = client.getWriteApi(
 
 setInterval(() => {
   writeClient.flush().catch(err => {
-    Sentry.captureException(err)
+    if (!/HttpError|getAddrInfo|RequestTimedOutError/i.test(String(err))) {
+      Sentry.captureException(err)
+    }
```
bajtos, line 27: Have you considered using a reverse logic & early return instead? It will make your change smaller.

```suggestion
    if (/HttpError|getAddrInfo|RequestTimedOutError/i.test(String(err))) return;
    Sentry.captureException(err)
```

It would be also nice to add a code comment to explain why are we ignoring these errors.

bajtos (approved):
Looks mostly good to me.

Do you have any concerns about ignoring too many errors? What if the upcoming wallet API integration fails with `HttpError`, would your pull request silently discard it too? Do we want that?

juliangruber:
> Do you have any concerns about ignoring too many errors? What if the upcoming wallet API integration fails with HttpError, would your pull request silently discard it too? Do we want that?

This code only affects errors thrown by the telemetry client itself, not errors that we submit to Sentry. So I think it should be fine! I'll add a comment for this

juliangruber:
Thank you for the review @bajtos!

PR #407, opened 2023-01-16
Hide parts of destination address

jopedroliveira:
- Hide the internal station address
- Add button to copy internal station address to clipboard
- Hide the destination address

### DIFF
--- a/renderer/src/components/FilAddressForm.tsx
+++ b/renderer/src/components/FilAddressForm.tsx
@@ -27,7 +27,7 @@ const FilAddressForm: FC<FilAddressFormProps> = ({ destinationAddress = '', save
       ref.current.focus()
       if (destinationAddress.length > 0) { ref.current.setSelectionRange(destinationAddress.length, destinationAddress.length) }
     }
-  // eslint-disable-next-line react-hooks/exhaustive-deps
+    // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [internalEditMode])
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


renderer/src/components/FilAddressForm.tsx
```diff
@@ -88,16 +88,25 @@ const FilAddressForm: FC<FilAddressFormProps> = ({ destinationAddress = '', save
...
           <span className='text-white hidden group-hover:block opacity-80 not-italic text-body-m font-body leading-[1.5rem]'>Edit</span>
         </button>
       </>
     )
   }
-
+  const renderDisplayAddress = () => {
+    if (!internalEditMode) {
+      if (inputAddr.length > 5) {
+        return `${inputAddr.slice(0, 6)} . . . ${inputAddr.slice(inputAddr.length - 6, inputAddr.length)}`
```
juliangruber, line 103: @bajtos proposed using `&hellip;` for the 3 points, what do you think about that? It's shorter and semantically correct

juliangruber (changes requested):

renderer/src/components/FilAddressForm.tsx
jopedroliveira: For this particular case, there's a space between each dot. I can replace it with 3x `&period;`, the full stop character, if you prefer the HTML code

juliangruber: It's not about the HTML code. If it's a design decision to have spaces between the dots, let's keep it as is

juliangruber (approved):

PR #419, opened 2023-01-18
limit max-len to 80 chars for non-jsx files

juliangruber:
I have noticed that often in PR reviews I would like to request a shorter line length, but the scalable way to prevent this is through linting instead.

A line length limit doesn't work well for JSX files, because especially css class names will often be too many, and breaking them into multi-line strings doesn't improve readability. Instead, I'm planning to follow up with another PR, that will add a rule like only one JSX element per line, or something similar.

I haven't yet fixed all files, but will do so if this PR gets approval.


build/download-saturn-l2.js
```diff
@@ -54,25 +55,33 @@ async function main () {
 
         if (res.status >= 300) {
           throw new Error(
-            `Cannot fetch saturn-l2 binary ${name}: ${res.status}\n${await res.text()}`
+            `Cannot fetch saturn-l2 binary ${name}: ${res.status}\n` +
+            await res.text()
```
juliangruber, line 59: We could also ignore template literals, check out the first commit for that version

build/download-saturn-l2.js
```diff
@@ -36,7 +36,8 @@ async function main () {
   await Promise.all(
     assets
       .map(async ({ name, browser_download_url: url }) => {
-        const match = name.match(/^L2-node_([A-Za-z0-9]+)_([A-Za-z0-9_]+)\.(tar\.gz|zip)$/)
+        const reg = /^L2-node_([A-Za-z0-9]+)_([A-Za-z0-9_]+)\.(tar\.gz|zip)$/
+        const match = name.match(reg)
```
bajtos, line 40: Coming from Prettier background, I would format this differently:

```js
        const match = name.match(
          /^L2-node_([A-Za-z0-9]+)_([A-Za-z0-9_]+)\.(tar\.gz|zip)$/
        )
```

build/download-saturn-l2.js
```diff
@@ -67,12 +68,17 @@ async function main () {
           await pipeline(res.body, gunzip(), tar.extract(outFile))
         } else {
           // Darwin needs to be a zip for notarization
-          await mkdir(path.join(outDir, 'l2node-darwin-x64'), { recursive: true })
+          await mkdir(
+            path.join(outDir, 'l2node-darwin-x64'),
+            { recursive: true }
+          )
```
bajtos, line 77: Similarly here:

```js
          await mkdir(path.join(outDir, 'l2node-darwin-x64'), {
            recursive: true 
          })
```

build/download-saturn-l2.js
```diff
@@ -67,12 +68,17 @@ async function main () {
...
           const parser = unzip.Parse()
           await Promise.all([
             (async () => {
               while (true) {
-                const [entry] = /** @type {[UnzipStreamEntry]} */ (await once(parser, 'entry'))
+                const [entry] =
+                  /** @type {[UnzipStreamEntry]} */
+                  await once(parser, 'entry')
```
bajtos, line 84: Are you sure you can remove those `()` characters? I think they are a necessary part of the syntax to tell TypeScript to cast the value to a different type.

How do you feel about this variant?

```js
                const [entry] = /** @type {[UnzipStreamEntry]} */ (
                  await once(parser, 'entry')
                )
```

build/download-saturn-l2.js
bajtos: I prefer to add the `+` sign at the next line. That way adding a new line or removing the last line does not require changes in other lines.

I think it's also common to indent the lines after the first one by one level.

```js
            `Cannot fetch saturn-l2 binary ${name}: ${res.status}\n`
              + await res.text()
```

IIRC, Prettier would format this line like this - placing the line break inside the `${}` expression. I find the result a bit difficult to read, but otherwise it has many nice properties.

```js
            `Cannot fetch saturn-l2 binary ${name}: ${res.status}\n${
              await res.text()
            }`
```

bajtos (approved):
LGTM.

It's a great idea to keep our lines reasonably long. 👍🏻 

While 80 characters is a common limit, I think nowadays 100 or even 120 is acceptable too. IIRC, GitHub can show up to 120 characters in diffs and other places.

I also have few suggestions for alternative ways how to split the long lines.

I don't want us to spend too much time on this, as I think it's not super important. Please make the decisions yourself, accept the suggestions you like and ignore the rest.

build/download-saturn-l2.js
juliangruber: Screenshot 2023-02-08 at 23 17 45

This is currently prevented, let's tackle it in a separate PR

main/preload.js
```diff
@@ -3,6 +3,7 @@
 /** @typedef {import('electron').IpcRendererEvent} IpcRendererEvent */
 /** @typedef {import('./typings').Activity} Activity */
 /** @typedef {import('./typings').FILTransaction} FILTransaction */
+// eslint-disable-next-line max-len
```
juliangruber, line 6: can you think of a way to improve these? Can typedef / type imports be written over multiple lines, or shorter?

build/download-saturn-l2.js
bajtos: > let's tackle it in a separate PR

👍🏻

main/preload.js
bajtos: Would something like this work?

```js
/** @typedef {
  import('./typings').FILTransactionProcessing
} FILTransactionProcessing */

File: main/telemetry.js [hunk 6]
```diff
@@ -11,7 +11,9 @@ const Sentry = require('@sentry/node')
 const client = new InfluxDB({
   url: 'https://eu-central-1-1.aws.cloud2.influxdata.com',
   // station-anonymous-write
-  token: '0fZyu9zjDvYlaNfOeuwgnQoUI0VcSzeYDpnOLjQyr30mz-Plqels5JHEwgKRbtCcDJbQmv62VnOV_FsZVxgoow=='
+  token:
+    '0fZyu9zjDvYlaNfOeuwgnQoUI0VcSzeYDpnOLjQyr30mz-Plqels5JHEwgKRbtCcDJbQmv62' +
+      'VnOV_FsZVxgoow=='
```
bajtos, line 16: Uh oh, splitting a magic string like this into two lines will make this code difficult to change when we rotate the token.

Let's ignore the max-len rule on this line please.

main/preload.js
juliangruber: it does! 🥳

PR #479, opened 2023-02-09
Switch from glif to filfox

juliangruber:
- Filfox pretty much gives us exactly the data we want, timely and complete
- Glif's data retention is problematic for the wallet. Transactions from some hours ago might not be returned by the API any more
- Switching from one dependency to another, FilFox is also well respected
- This simplifies everything so much


juliangruber:
hindsight please 🙏

main/wallet-backend.js
```diff
@@ -10,21 +10,15 @@ const { CoinType } = require('@glif/filecoin-address')
 const { strict: assert } = require('node:assert')
 const { Message } = require('@glif/filecoin-message')
 const { FilecoinNumber, BigNumber } = require('@glif/filecoin-number')
-const { request, gql } = require('graphql-request')
-const pMap = require('p-map')
+const { default: fetch } = require('node-fetch')
```
bajtos, line 13: Nitpick - feel free to ignore it.

I personally prefer to use `fetch` from `undici`. It's maintained by the Node.js project and shipped as the default `fetch` implementation from Node.js version 18.

After we upgrade to Electron 23, which ships Node.js 18 (https://github.com/filecoin-station/filecoin-station/pull/476), we should be able to use this built-in `fetch` implementation and can remove dependency on `undici` or `node-fetch`.

Maybe you can add a code comment or create a GH issue to remind us to start using the built-in `fetch` implementation after upgrading to Node.js 18?

main/wallet-backend.js
```diff
@@ -182,98 +173,15 @@ class WalletBackend {
...
-      }
-    `
-    const variables = {
-      address,
-      limit: 100,
-      offset: 0
-    }
-    /** @type {{messages: GQLMessage[] | null}} */
-    const { messages } = await request(this.url, query, variables)
+    const url = `https://filfox.info/api/v1/address/${address}/messages?pageSize=100`
```
bajtos, line 181: IIUC, you are always fetching only the first 100 messages. Could this become a problem? E.g. if our wallet shows ALL transactions, we will need to load more messages beyond the first 100 progressively. (E.g. add a "load more" button in the UI that will fetch the subsequent 100 messages.)

The hard-coded limit of 100 messages is acceptable for now, but I would consider opening a follow-up GH issue if we think this is something to improve in the future.

It would also be good to show some sort of a visual indicator to let the user know when the list of transactions was truncated. I think this has a relatively higher priority.

bajtos (commented):
I love how much simpler our code is after this change 👍🏻

main/wallet-backend.js
juliangruber: Fix in https://github.com/filecoin-station/filecoin-station/commit/23c1d9cc8cd7e1f40938500e60dfc1582817692f 😁

bajtos: > Fix in [23c1d9c](https://github.com/filecoin-station/filecoin-station/commit/23c1d9cc8cd7e1f40938500e60dfc1582817692f) 😁

LOL 🤣

PR #521, opened 2023-02-20
Display tray on/off state

juliangruber:
This is a follow-up on https://github.com/filecoin-station/filecoin-station/pull/177, where we added the tray assets, but the logic wasn't yet wired up.

This makes it so the Station tray icon changes depending on whether you're connected to the Saturn network.

### DIFF
--- a/main/tray.js
+++ b/main/tray.js
@@ -95,9 +95,8 @@ function setupIpcEventListeners (contextMenu, ctx) {
   ipcMain.on(ipcMainEvents.ACTIVITY_LOGGED, () => {
     assert(tray)
 
-    const entries = ctx.getAllActivities()
     let connected = false
-    for (const entry of entries.reverse()) {
+    for (const entry of ctx.getAllActivities().reverse()) {
       if (
         entry.source === 'Saturn' &&
         entry.type === 'info' &&
--- a/main/tray.js
+++ b/main/tray.js
@@ -25,7 +25,6 @@ module.exports = function (/** @type {Context} */ ctx) {
   const image = nativeImage.createFromPath(icon('off'))
   image.setTemplateImage(true)
   tray = new Tray(image)
-
   const contextMenu = Menu.buildFromTemplate([
     {
       label: `Filecoin Station v${STATION_VERSION}`,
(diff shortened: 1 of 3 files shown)


main/tray.js
```diff
@@ -85,6 +91,40 @@ function setupIpcEventListeners (contextMenu) {
...
+            entry.message === 'Saturn Node started.' ||
+            entry.message.includes('was able to connect') ||
+            entry.message.includes('will try to connect')
+          )
+        )
+      ) {
+        connected = false
+        break
+      }
+    }
```
bajtos, line 120: I find it a bit hacky and potentially brittle to detect online/offline mode here in the tray module. Have you considered moving this code into the saturn module?

For example:
- The saturn module (and each new module in the future) should keep track whether it's online or offline, potentially using this activity-log-based solution. The status can be exposed as a property, e.g. `saturn.isOnline` or maybe `saturn.status` that can be `'online' | 'offline' | 'connecting'`.
- Here in the tray, we can keep listening for ACTIVITY_LOGGEd events and use the `saturn.status` variable to decide whether we are online or not. In the future, when we add more modules, we can extend this check to combine status from all modules.

juliangruber: good idea to consolidate the concerns, will do!

main/tray.js
```diff
@@ -85,6 +91,15 @@ function setupIpcEventListeners (contextMenu) {
     getItemById('checkingForUpdates').visible = false
   })
 
+  ipcMain.on(ipcMainEvents.ACTIVITY_LOGGED, () => {
+    assert(tray)
+
+    const state = saturn.isOnline() ? 'on' : 'off'
+    const image = nativeImage.createFromPath(icon(state))
+    image.setTemplateImage(true)
+    tray.setImage(image)
```
bajtos, line 100: How expensive is it to create a new native image? Can we create these two images (on & off) at start time, keep them around for the entire life of the app, and then just select the right image based on the state in this code?

Something along the following lines:

```js
    const trayIcon = saturn.isOnline() ? iconForStateOn : iconForStateOff;
    tray.setImage(trayIcon)
```

bajtos (approved):
Much better!

I have one performance-related concern; see below. 

Other than that, the PR is good to be merged.

main/tray.js
juliangruber: great suggestion!

juliangruber:
follow-up PR: https://github.com/filecoin-station/filecoin-station/pull/532

bajtos (approved):

PR #533, opened 2023-02-23
set project max line length to 115

juliangruber:
115 characters is the maximum line length (for me) that GitHub can display without scrolling horizontally.


bajtos (approved):
Some people argue for using 80 chars even if GitHub can show more. I was personally using 100 as a middle ground, but I don't have strong opinions here. If you find 115 a reasonable limit, I am fine with that.

juliangruber:
I personally prefer 80 chars still, and I'm trying to get to that as close as possible, in reasonable steps that we can agree on 🤞

PR #560, opened 2023-03-06
Fix saturn connectivity

juliangruber:
I haven't looked into the specifics yet, but with most recent nodejs and electron versions, `localhost` doesn't work any more. Fix for https://github.com/filecoin-station/desktop/pull/558

### DIFF
--- a/main/saturn-node.js
+++ b/main/saturn-node.js
@@ -140,7 +140,7 @@ async function start (/** @type {Context} */ ctx) {
 
       const apiMatch = output.match(/^API: (http.*)$/m)
       if (apiMatch) {
-        apiUrl = apiMatch[1]
+        apiUrl = apiMatch[1].replace('localhost', '127.0.0.1')
 
         appendToChildLog('Saturn node is up and ready')
         console.log('Saturn node is up and ready (API URL: %s)', apiUrl)


main/saturn-node.js
```diff
@@ -140,7 +140,7 @@ async function start (/** @type {Context} */ ctx) {
 
       const apiMatch = output.match(/^API: (http.*)$/m)
       if (apiMatch) {
-        apiUrl = apiMatch[1]
+        apiUrl = apiMatch[1].replace('localhost', '127.0.0.1')
```
bajtos, line 143: Would it be too difficult to make this change in the L2-node module instead? 

https://github.com/filecoin-saturn/L2-node/blob/ae78fe055f49a9c97eb099bf1ccb74e2e23c0ef7/cmd/saturn-l2/main.go#L335-L337

```go
	log.Infof("Server listening on %v", nl.Addr())
	fmt.Printf("WebUI: http://localhost:%d/webui\n", port)
	fmt.Printf("API: http://localhost:%d/\n", port)
```

juliangruber: It wouldn't be, it's just more work to also release it and at this point the project is basically unmaintained :|

bajtos: Well, the release process is super simple. You just create a new tag and then the CI/CD pipeline takes care of the rest.

https://github.com/filecoin-saturn/L2-node#release-process

Anyhow, as long as this one-off fix is inside `saturn-node.js` only and does not proliferate into other (future) modules, then I am fine with that.

bajtos (approved):

main/saturn-node.js
juliangruber: Ok, sounds good!

PR #599, opened 2023-03-24
telemetry: add periodic `ping`, independent of jobs

juliangruber:
Previously for every job completed, we were also submitting machine info like wallet hash, operating system, etc. This has two problems:
1. When there are no jobs being completed (eg Saturn testnet traffic problem), we don't know how many Stations there are
2. There is a lot of redundant information submitted.

In this PR I'm introducing a new point `ping`, with all that meta attached, which gets submitted per Station every 10s. This can also be lowered to every 1m if we feel it's too much. In turn, the extra metadata has been removed from `jobs-completed`.

Through this, we can still construct all the queries from our stats dashboard. Some information is lost, but we hadn't used it anyway, like: how many jobs are completed per platform. If we want to add these back, we can do that, but I propose we do it in addition to the periodic `ping`.

If and once this is merged, port to `core`.

### DIFF
--- a/main/telemetry.js
+++ b/main/telemetry.js
@@ -16,7 +16,7 @@ const client = new InfluxDB({
     '0fZyu9zjDvYlaNfOeuwgnQoUI0VcSzeYDpnOLjQyr30mz-Plqels5JHEwgKRbtCcDJbQmv62VnOV_FsZVxgoow=='
 })
 
-export const writeClient = client.getWriteApi(
+const writeClient = client.getWriteApi(
   'Filecoin Station', // org
   'station', // bucket
   'ns' // precision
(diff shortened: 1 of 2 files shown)


main/telemetry.js
```diff
@@ -25,41 +23,36 @@ const writeClient = client.getWriteApi(
...
 setInterval(() => {
-  writeClient.flush().catch(err => {
-    // Ignore unactionable InfluxDB errors
-    // eslint-disable-next-line max-len
-    const reg = /HttpError|getAddrInfo|RequestTimedOutError|ECONNRESET|CERT_NOT_YET_VALID/i
-    if (!reg.test(String(err))) {
-      Sentry.captureException(err)
-    }
-  })
-}, 5000).unref()
```
bajtos, line 36: Would you mind shuffling the new code around to let `git diff` understand that we are not making any changes here?

bajtos (approved):

main/telemetry.js
juliangruber: ok, will try!

Issue #35, opened 2022-06-07
Support Apple Silicon M1/M2 (darwin arm64)


bajtos:
Improve our build & release setup to produce binaries for `darwin-arm64` platform in addition to `darwin-x86_64`.

Related info:
- https://github.com/electron-userland/electron-builder/pull/5426#issuecomment-749684326
- Example PR adding M1 support to a different project: https://github.com/athensresearch/athens/pull/707

Consider building a universal binary that supports both architectures - see https://github.com/electron/universal

bajtos:
I think we will need to update saturn-l2 release process to notarise the arm64 binary.

See https://github.com/goreleaser/goreleaser/issues/1227 and https://goreleaser.com/customization/sign/?h=gon#with-gon

bajtos:
I opened an issue to fix saturn-l2 darwin-arm64 builds here: https://github.com/filecoin-project/saturn-l2/issues/35

juliangruber:
Won't happen in this sprint, as I'm having apple credential issues, and we don't yet have our own credentials

PR #630, opened 2023-04-11
feat: report arch=arm64 in Rosetta 2

bajtos:
`os.arch()` returns `x64` when running in a translated environment on ARM64 machine, e.g. via macOS Rosetta. We don't provide Apple `arm64` builds yet.  As a result, Stations running on Apple Silicon were reporting `x64` arch.  To fix the problem, we detect `arm64` translation and report a different architecture.

## Background

I am running Station Desktop version 0.15.3 on an Apple M1 machine. Yet when I look at our Graphana Dashboard, I see that all stations run on (Intel) x64 architecture.

In the electron log file `~/Library/Logs/Filecoin Station/main.log`, I see the following lines:

```
[info]  (main)    Filecoin Station build version: 0.15.3 darwin-x64
[info]  (main)    Machine spec: Darwin version 22.4.0
```

Source code printing these lines:

https://github.com/filecoin-station/desktop/blob/175e1c4b97d4af982efa8b048d588bb4f530e468/main/index.js#L48-L56

It seems that `os.arch()` is returning `x64` even though we are running on `arm64`.

I checked the problem using x64 version of Node.js executed on my arm64 machine:

```
❯ ~/Downloads/node-v18.15.0-darwin-x64/bin/node -e "console.log(os.arch())"
x64
```

Using the arm64 version of Node:

```
❯ node -e "console.log(os.arch())"
arm64
```

This seems to be expected behaviour, see
- https://apple.stackexchange.com/q/427970/51077
- https://stackoverflow.com/q/65346260/69868
- https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment#3616845

### DIFF
--- a/main/index.js
+++ b/main/index.js
@@ -59,6 +59,12 @@ log.info('Machine spec: %s version %s', os.type(), os.release())
 // log.info('Machine spec: %s %s version %s', os.type(),
 //   os.machine(),
 //   os.release())
+if (app.runningUnderARM64Translation) {
+  log.warn(
+    'Warning: we are running under ARM64 translation' +
+      ' (macOS Rosetta or Windows WOW).'
+  )
+}
 
 // Expose additional metadata for Electron preload script
 process.env.STATION_BUILD_VERSION = BUILD_VERSION
--- a/main/telemetry.js
+++ b/main/telemetry.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const { app } = require('electron')
 const { InfluxDB, Point } = require('@influxdata/influxdb-client')
 const { createHash } = require('node:crypto')
 const { getDestinationWalletAddress } = require('./station-config')
@@ -49,7 +50,12 @@ setInterval(() => {
   point.stringField('version', pkg.version)
   point.tag('station', 'desktop')
   point.tag('platform', platform())
-  point.tag('arch', arch())
+  // `os.arch()` returns `x64` when running in a translated environment on ARM64
+  // machine, e.g. via macOS Rosetta. We don't provide Apple `arm64` builds yet.
+  // As a result, Stations running on Apple Silicon were reporting `x64` arch.
+  // To fix the problem, we detect `arm64` translation and report a different
+  // architecture.
+  point.tag('arch', app.runningUnderARM64Translation ? 'arm64' : arch())
   writeClient.writePoint(point)
 }, 10_000).unref()
 


main/telemetry.js
```diff
@@ -49,7 +50,12 @@ setInterval(() => {
   point.stringField('version', pkg.version)
   point.tag('station', 'desktop')
   point.tag('platform', platform())
-  point.tag('arch', arch())
+  // `os.arch()` returns `x64` when running in a translated environment on ARM64
+  // machine, e.g. via macOS Rosetta. We don't provide Apple `arm64` builds yet.
+  // As a result, Stations running on Apple Silicon were reporting `x64` arch.
+  // To fix the problem, we detect `arm64` translation and report a different
+  // architecture.
+  point.tag('arch', app.runningUnderARM64Translation ? 'arm64' : arch())
```
bajtos, line 58: Discussion point: do we want to report both the build-arch (x64) and machine-arch (arm64) to measure how many people are running a version built for a different architecture?

It does not matter right now (we provided only x64 builds for macOS, all Macs are running the emulation), but it may become interesting after we implement support for Apple Silicon - #35.

bajtos:
How to debug this problem:

- When you run `npm start` from a native arm64 shell, Electron reports `arm64` arch.
- You need to package the app first and then run the artifact produced by electron-build

```
❯ npm run package
(...)
❯ ./dist/mac/Filecoin\ Station.app/Contents/MacOS/Filecoin\ Station
```

Screenshot 2023-04-11 at 17 05 48

main/telemetry.js
juliangruber: This is a cool idea for the future, hopefully we'll add native builds before we need to implement it I think

main/index.js
```diff
@@ -59,6 +59,12 @@ log.info('Machine spec: %s version %s', os.type(), os.release())
 // log.info('Machine spec: %s %s version %s', os.type(),
 //   os.machine(),
 //   os.release())
+if (app.runningUnderARM64Translation) {
+  log.warn(
+    'Warning: we are running under ARM64 translation' +
+      ' (macOS Rosetta or Windows WOW).'
+  )
+}
```
juliangruber, line 67:
```suggestion
```

Do you think this log message is important? Our logs are quite noisy already, and we aren't doing anything in the electron process that is CPU intensive

juliangruber (approved):

main/index.js
bajtos: I added this log message to help us with troubleshooting. Maybe I can rephrase it from a warning to an informational message?

juliangruber: Ok I can see that it's useful now. Warning sgtm

PR #692, opened 2023-05-11
Fix more Core crashes

juliangruber:


main/core.js
```diff
@@ -15,14 +15,13 @@ const { Core } = require('@filecoin-station/core')
...
-  __dirname,
-  '..',
+  app.getAppPath(),
   'node_modules',
   '@filecoin-station',
   'core',
   'bin',
   'station.js'
-)
+).replace('app.asar', 'app.asar.unpacked')
```
bajtos, line 24: I am confused. Why are we both unpacking asar and adding `node_modules/@filecoin-station/core/modules` to extra resources? As I understand how electron works, only one of those things should be necessary. Maybe my understanding is wrong. 

Could you please explain a bit more how does the new solution work? Maybe capture that description in a code comment or developer documentation.

bajtos (commented):
I like that you removed the `execa` dependency 👍🏻 

I don't fully understand the fix, though; see my comments below.

main/tray.js
```diff
@@ -114,6 +114,7 @@ function setupIpcEventListeners (contextMenu, ctx) {
 
   ipcMain.on(ipcMainEvents.ACTIVITY_LOGGED, updateTray)
   ipcMain.on(ipcMainEvents.UPDATE_AVAILABLE, updateTray)
+  setInterval(updateTray, 1000)
```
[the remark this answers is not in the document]
juliangruber: I should have removed it, I don't think we need the change really

main/core.js
juliangruber: This is one of the advised ways to target files either from the working directory or `app.asar.unpacked`. See https://github.com/electron/electron/issues/6262#issuecomment-229043051

We are unpacking asar in order to get direct access to the files. We add `node_modules...` because that's where we unpacked the files to:

```bash
$ ls dist/mac/Filecoin\ Station.app/Contents/Resources/app.asar.unpacked/node_modules/@filecoin-station/core/
```

Would you prefer to unpack the files somewhere else? I like that the location inside `app.asar.unpacked` is the same as inside `app.asar`.

Wdyt?

juliangruber: For docs, do you think linking to this comment is enough? To me personally it sufficiently explains the situation

bajtos: > This is one of the advised ways to target files either from the working directory or `app.asar.unpacked`. See [electron/electron#6262 (comment)](https://github.com/electron/electron/issues/6262#issuecomment-229043051)
> 
> We are unpacking asar in order to get direct access to the files. We add `node_modules...` because that's where we unpacked the files to:
> 
> ```shell
> $ ls dist/mac/Filecoin\ Station.app/Contents/Resources/app.asar.unpacked/node_modules/@filecoin-station/core/
> ```
> 
> Would you prefer to unpack the files somewhere else? I like that the location inside `app.asar.unpacked` is the same as inside `app.asar`.

I understand the part about getting access to unpacked files. What was not clear to me is 1) how is electron-builder determining that Core should be unpacked 2) how is the build config `extraResources` related to that

I removed `extraResources` from `electron-builder.yaml` and AFAICT, all works well. I added the commit 1377183 to your PR. Please take a look and feel free to revert it.

> For docs, do you think linking to this comment is enough? To me personally it sufficiently explains the situation

I would prefer to add a short comment before `const corePath` line to explain that the next line will get a path pointing into the ASAR archive, and we need to change it to point to the unpacked archive directory, and that electron-builder is smart enough to figure out that it needs to unpack Core and its dependencies.

bajtos: Related Electron & electron-builder docs:

### extraResources[¶](https://www.electron.build/configuration/contents#extraresources)

Array<String | FileSet> | String | FileSet

A [glob patterns](https://www.electron.build/file-patterns) relative to the project directory, when specified, copy the file or directory with matching names directly into the app’s resources directory (Contents/Resources for MacOS, resources for Linux and Windows).

File patterns (and support for from and to fields) the same as for [files](https://www.electron.build/configuration/contents#files).

### extraFiles[¶](https://www.electron.build/configuration/contents#extrafiles)

Array<String | FileSet> | String | FileSet

The same as [extraResources](https://www.electron.build/configuration/contents#extraresources) but copy into the app’s content directory (Contents for MacOS, root directory for Linux and Windows).

### Overridable per Platform Options[¶](https://www.electron.build/configuration/configuration#overridable-per-platform-options)

- `extraResources` The [extra resources](https://www.electron.build/configuration/contents#extraresources) configuration.
- `extraFiles` The [extra files](https://www.electron.build/configuration/contents#extrafiles) configuration.
- `asar` = true [AsarOptions](https://www.electron.build/configuration/configuration#AsarOptions) | Boolean | “undefined” - Whether to package the application’s source code into an archive, using [Electron’s archive format](http://electron.atom.io/docs/tutorial/application-packaging/).
Node modules, that must be unpacked, will be detected automatically, you don’t need to explicitly set [asarUnpack](https://www.electron.build/configuration/configuration#configuration-asarUnpack) - please file an issue if this doesn’t work.
- `asarUnpack` Array<String> | String | “undefined” - A [glob patterns](https://www.electron.build/file-patterns) relative to the [app directory](https://www.electron.build/configuration/configuration#MetadataDirectories-app), which specifies which files to unpack when creating the [asar](http://electron.atom.io/docs/tutorial/application-packaging/) archive.

bajtos (approved):
LGTM after I removed `extraResources` config that seems to be not needed.

If you find that it's needed, then the previous version with `extraResources` is fine with me too.

Please consider adding a code comment as I suggested in my comment below.

main/core.js
juliangruber: 1ccbc52

electron-builder.yml
```diff
@@ -12,6 +12,11 @@ files:
     - node_modules/**/*
     - package.json
 
+extraResources:
+  - from: 'node_modules/@filecoin-station/core/modules'
+    to: 'node_modules/@filecoin-station/core/modules'
+    filter: ['*']
+
 asarUnpack: 'main/**/scripts/**/*'
```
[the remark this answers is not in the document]
juliangruber: https://github.com/filecoin-station/desktop/pull/693

juliangruber:
https://github.com/filecoin-station/desktop/releases/tag/v0.16.0-3

PR #726, opened 2023-05-26
integrate core 11

juliangruber:


juliangruber:
@bajtos hindsight please

juliangruber:
https://github.com/filecoin-station/desktop/releases/tag/v0.17.0

main/core.js
```diff
@@ -39,86 +37,62 @@ async function setup (/** @type {Context} */ ctx) {
...
+    },
+    stdio: ['pipe', 'pipe', 'pipe', 'ipc']
+  })
+
+  assert(childProcess.stdout)
+  childProcess.stdout.setEncoding('utf8')
+  childProcess.stdout.on('data', chunk => {
+    logs.push(chunk)
+    for (const line of chunk.split('\n').filter(Boolean)) {
+      const event = JSON.parse(line)
```
bajtos, line 70: Is it safe to assume that inter-process stdio will always split the data chunks on new-line boundary? If not, then we may receive a single line split across two or more chunks.

I am proposing to pipe stdio streams to `split2` before we parse it as ND-JSON. See https://github.com/mcollina/split2

main/core.js
```diff
@@ -39,86 +37,62 @@ async function setup (/** @type {Context} */ ctx) {
...
+            ...event,
+            type: event.type.replace('activity:', ''),
+            timestamp: new Date(),
+            id: randomUUID()
+          }
+          activities.push(ctx, activity)
+          break
+        }
+        default:
+          throw new Error(`Unknown event type: ${event.type}`)
```
bajtos, line 87: I think this is going to crash the entire process on unhandled error, is that correct? That seems to be too strict to me. Have you considered treating this edge case as a warning only? I.e. log this error to console.error and send it to Sentry.

main/core.js
```diff
@@ -130,10 +104,9 @@ async function start (core) {
     console.log(`Core closed all stdio with code ${code ?? '<no code>'}`)
 
     ;(async () => {
-      const log = await core.logs.get()
       Sentry.captureException('Core exited', scope => {
         // Sentry UI can't show the full 100 lines
-        scope.setExtra('logs', log.split('\n').slice(-10).join('\n'))
+        scope.setExtra('logs', logs.getLast(10))
```
bajtos, line 109: Before this change, you added last 10 lines of logs. Now you are adding last 10 chunks. I am not sure if chunks are always split on line boundaries. Maybe we should pipe `childProcess.stderr` through `split2` before we read the chunks into our `log` array?

main/index.js
```diff
@@ -97,32 +95,14 @@ app.on('second-instance', () => {
...
 })
 
-if (isDev) {
-  // Do not preserve old Activity entries in development mode
-  (async () => {
-    try {
-      await fs.writeFile(await getActivityFilePath(), '')
-    } catch {}
-  })()
-}
```
bajtos, line 107: Could you please remind me what is the new behaviour? IIUC, we agreed that Activity entries are no longer persisted, is that correct? If that's the case, then we don't need special handling for the development mode, and this removal is fine 👍🏻

bajtos (commented):
Awesome 👍🏻 

I have few comments to consider, PTAL.

main/index.js
juliangruber: This is correct 👍

main/core.js
juliangruber: `Logs#push` performs splitting by newline, doesn't that make it work?

main/core.js
juliangruber: Good idea! https://github.com/filecoin-station/desktop/pull/730

main/core.js
juliangruber: good idea! https://github.com/filecoin-station/desktop/pull/731

main/core.js
bajtos: What if the child process send a long line in two chunks? The first chunk does not contain any newline character, the second one does. Splitting on newlines will keep two lines in the `logs` array. Unless you join all items first and then split the combined string on newlines. (That feels very inefficient to me, TBH.)

Anyhow, not a big deal, I think this can produce an annoyance at most, it won't be a major problem.

juliangruber: I see! The `split2` PR will help us make sure this is not an issue.

PR #730, opened 2023-05-29
core: fix too strict error handling

juliangruber:


main/core.js
```diff
@@ -67,7 +67,13 @@ async function start (ctx) {
   childProcess.stdout.on('data', chunk => {
     logs.push(chunk)
     for (const line of chunk.split('\n').filter(Boolean)) {
-      const event = JSON.parse(line)
+      let event
+      try {
+        event = JSON.parse(line)
+      } catch (err) {
+        console.error(err)
```
bajtos, line 74: `JSON.parse` returns very cryptic errors. We should add more context to the error message.


```suggestion
        err.message = `Cannot parse line from Station Core stdout: ${err.message}`
        console.error(err)
```

```

**Review by bajtos — CHANGES_REQUESTED** [review]

**Review by bajtos — CHANGES_REQUESTED** [review]
The changes are going in the right direction. I am proposing to improve few details though.

File: main/core.js (hunk 1 above)
Comment (juliangruber) [reply main/core.js:74]: I didn't improve these messages because we can see stack traces in Sentry (telling us where an error comes from), but it's nice this way nevertheless

Comment (juliangruber) [reply main/core.js:74]: Screenshot 2023-05-30 at 15 13 13

Do you know how to fix this ts error?

Comment (bajtos) [reply main/core.js:74]: Yeah, error handling in TypeScript is tricky because it forces you to deal with the fact that `err` can be anything.

Would something like this work?

```ts
const error = err instanceof Error ? err : new Error(String(err));
error.message = // ...
```

Another solution is to explicitly tell TypeScript that `err` is `any`. IIRC, this was not supported for JSDoc typings in the past, I don't remember if works nowadays.

```js
try {
} catch (/** @type {any} */ err) {
 // ...
}
```

bajtos: > According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse, the method can only throw SyntaxError, which is an error class. So I don't think TS is right here 🤔

TypeScript does not have a concept of a list of exceptions that a function can throw. Therefore it cannot infer that your `try` block can only throw SyntaxError.

bajtos: > I didn't improve these messages because we can see stack traces in Sentry (telling us where an error comes from),

Makes sense. I was thinking more about the situation when we are troubleshooting the app locally.

bajtos (approved):
👌🏻

main/core.js
juliangruber: Don't we also get stacktraces then, in the terminal?

PR #731, opened 2023-05-29
core: fix nldjson parsing using `split2`

juliangruber:


main/core.js
```diff
@@ -64,9 +65,10 @@ async function start (ctx) {
...
   assert(childProcess.stdout)
   childProcess.stdout.setEncoding('utf8')
-  childProcess.stdout.on('data', chunk => {
-    logs.push(chunk)
-    for (const line of chunk.split('\n').filter(Boolean)) {
+  childProcess.stdout
+    .pipe(split2())
+    .on('data', line => {
+      logs.push(line)
       const event = JSON.parse(line)
```
bajtos, line 72: I think we should wrap this line in a try/catch block so that a non-JSON line does not crash the process. WDYT?

bajtos (approved):
💪🏻

main/core.js
```diff
@@ -69,7 +69,15 @@ async function start (ctx) {
     .pipe(split2())
     .on('data', line => {
       logs.push(line)
-      const event = JSON.parse(line)
+      let event
+      try {
+        event = JSON.parse(line)
+      } catch {
+        const err = new Error(`Failed to parse core event: ${line}`)
```
bajtos, line 76: This will discard the stack trace from the original error. How about storing the original error in `err.cause`?

juliangruber: Yeah, but does that matter here? The original error is thrown 2 lines above, and the new error has a stack trace too

bajtos: Fair enough 👍🏻

Although I prefer to follow best practices everywhere instead of deciding whether we can make an exception on a case-by-case basis.

Should we at least include the JSON parser error in the message? ATM, we tell the user that JSON is malformed but don't explain where exactly.

juliangruber: Ah I see! I didn't realize `err.cause` is supported yet, so I didn't see why attaching the error would help. But now that I saw that it will also help include the original message, I'm going to change to that 👍

PR #751, opened 2023-06-02
activities: fix use spark not saturn

juliangruber:


main/activities.js
```diff
@@ -16,8 +16,8 @@ class Activities {
   push (ctx, activity) {
     this.#activities.push(activity)
     this.#activities.splice(0, this.#activities.length - 100)
-    ctx.recordActivity(activity)
     this.#detectChangeInOnlineStatus(activity)
+    ctx.recordActivity(activity)
```
juliangruber, line 20: needed to move this after `#detectChangeInOnlineStatus`, so that when the activity is emitted the online status is already correct. Otherwise we'd stay offline until the next activity is recorded. I thought the asynchronicity of IPC would make this irrelevant, but it does change that.

main/activities.js
```diff
@@ -26,14 +26,13 @@ class Activities {
   #detectChangeInOnlineStatus (activity) {
     if (
       activity.type === 'info' &&
-      activity.message.includes('Saturn Node is online')
+      (activity.message === 'SPARK started reporting retrievals' ||
+      activity.message === 'SPARK retrieval reporting resumed')
     ) {
+      console.log('XXX went online')
```
bajtos, line 32: Why `XXX` and not `SPARK`?

main/activities.js
bajtos: This is very subtle! Have you considered capturing this GH comment in a code comment to make this explanation easier to find in the future?

main/activities.js
juliangruber: ooops that was a debug call. will remove

juliangruber: c274852

PR #815, opened 2023-06-27
ci: add auto approve dependabot PRs

juliangruber:
For certain packages, there's nothing to do for us as long as tests pass.

The list of package names can be extended in the future to include more safe dependencies.

### DIFF
--- a/.github/workflows/dependabot-auto-approve.yml
+++ b/.github/workflows/dependabot-auto-approve.yml
@@ -15,7 +15,14 @@ jobs:
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
       - name: Approve a PR
-        if: ${{(contains(steps.metadata.outputs.dependency-names, 'eslint') || contains(steps.metadata.outputs.dependency-names, '@types')) && (steps.metadata.outputs.update-type == 'version-update:semver-patch' || steps.metadata.outputs.update-type == 'version-update:semver-minor')}}
+        if: >
+          ${{(
+            contains(steps.metadata.outputs.dependency-names, 'eslint') ||
+            contains(steps.metadata.outputs.dependency-names, '@types')
+          ) && (
+            steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
+            steps.metadata.outputs.update-type == 'version-update:semver-minor'
+          )}}
         run: gh pr review --approve "$PR_URL"
         env:
           PR_URL: ${{github.event.pull_request.html_url}}
--- a/.github/workflows/dependabot-auto-approve.yml
+++ b/.github/workflows/dependabot-auto-approve.yml
@@ -10,9 +10,10 @@ jobs:
     if: ${{ github.actor == 'dependabot[bot]' }}
     strategy:
       matrix:
-        dependency:
+        dependencyStartsWith:
           - eslint
-          - '@types'
+          - '@typescript-eslint/'
+          - '@types/'
     steps:
       - name: Dependabot metadata
         id: metadata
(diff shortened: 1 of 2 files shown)


.github/workflows/dependabot-auto-approve.yml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Approve a PR
+        if: ${{(contains(steps.metadata.outputs.dependency-names, 'eslint') || contains(steps.metadata.outputs.dependency-names, '@types')) && (steps.metadata.outputs.update-type == 'version-update:semver-patch' || steps.metadata.outputs.update-type == 'version-update:semver-minor')}}
```
bajtos, line 18: This condition is very long. Is there a way to split it into multiple lines? I think YAML does support multi-line values, although I don't know the syntax.

Would something like this work? (Based on https://stackoverflow.com/q/3790454/69868)

```yaml
if: >
  ${{(
    contains(steps.metadata.outputs.dependency-names, 'eslint') ||   
    contains(steps.metadata.outputs.dependency-names, '@types')
  ) && (
    steps.metadata.outputs.update-type == 'version-update:semver-patch' ||   
    steps.metadata.outputs.update-type == 'version-update:semver-minor'
  )}}
```

It would be even nicer if we could provide the list of pre-approved dependencies as an array, so that we don't have to repeat the part `contains(steps.metadata.outputs.dependency-names`. This is nice to have though.

Maybe we could auto-approve Dependabot PRs for all `devDependencies` specified in `package.json`?

Anyhow, it's probably best to start with what you have and then improve this initial solution incrementally over time.

bajtos, line 18: Is it safe to use `contains`? Have you considered using [startsWith](https://docs.github.com/en/actions/learn-github-actions/expressions#startswith) for a more strict check?

bajtos (approved):
Nice!

I have a few suggestions for improving the long `if` condition, but the current solution is also good.

.github/workflows/dependabot-auto-approve.yml
juliangruber: We could make it more precise but need to grow it then. Currently it captures these dependencies:

Screenshot 2023-06-28 at 10 18 07

Wdyt?

juliangruber: Good suggestion with the multiline string. I had tried it but couldn't figure out the right syntax. 🙏 

I've also added a `matrix` so that we configure a list of dependencies once.

> Maybe we could auto-approve Dependabot PRs for all devDependencies specified in package.json?

I think this is too wide a change. For example with `electron` we still need manual review. So far I've only added the linter and ts utilities, which for minor/patch changes are safe to me.

bajtos: I would create two `startsWith` conditions: `@typescript-eslint` and `eslint`.

Having said that, I think the likelihood of a production dependency containing `eslint` in the name is low enough to make `contains` a good enough solution.

bajtos: > I've also added a `matrix` so that we configure a list of dependencies once.

I don't see this change in the diff. Did you perhaps forget to push some commits?

.github/workflows/dependabot-auto-approve.yml
```diff
@@ -0,0 +1,29 @@
...
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Approve a PR
+        if: >
+          ${{(
+            contains(steps.metadata.outputs.dependency-names, 'eslint') ||
+            contains(steps.metadata.outputs.dependency-names, '@types')
+          ) && (
+            steps.metadata.outputs.update-type == 'version-update:semver-patch' ||
+            steps.metadata.outputs.update-type == 'version-update:semver-minor'
+          )}}
```
bajtos, line 25: Nice! This is much easier to read 👍🏻

bajtos (approved):
:shipit:

.github/workflows/dependabot-auto-approve.yml
juliangruber: Good point! I'll refactor it. Better safe than sorry

PR #931, opened 2023-08-25
feat: automatically restart to update

bajtos:
When the Station is running in the tray, and a new version is detected, we immediately trigger "restart to install".

The goal is to speed up the adoption of new versions of Station Desktop.

Screenshot of the notification:

Screenshot 2023-08-25 at 15 28 15

How I tested the change:
1. `npm run package`
2. `open dist/Filecoin\ Station.dmg`
3. Check that the new notification is shown
4. Check that `$HOME/Library/Logs/Filecoin Station/main.log` contains the new message about restart:
    ```
    Restarting Station to install the new version
    ```

Unfortunately, the auto-update failed while verifying the code signature of the new version. I think this could be caused by running a locally-built Station and trying to update to one built by our CI.


juliangruber:
> Unfortunately, the auto-update failed while verifying the code signature of the new version. I think this could be caused by running a locally-built Station and trying to update to one built by our CI.

That's it. To test it properly we'd have to create a release. I think we can skip this for now

main/ui.js
```diff
@@ -35,6 +35,9 @@ module.exports = async function (ctx) {
     }
   })
 
+  ui.on('show', () => { ctx.isShowingUI = true })
+  ui.on('hide', () => { ctx.isShowingUI = false })
```
juliangruber, line 39: Could we use a simpler https://www.electronjs.org/docs/latest/api/browser-window#winisvisible instead?

bajtos: I am not sure if that's the same thing. Quoting from the docs:

> Returns boolean - Whether the window is visible to the user in the foreground of the app.


What if the Station's window is in the background, underneath windows of other applications? What if it's minimised?

Is it worth investing our time to better understand when `win.isVisible()` returns true/false; or is the current solution good enough?

bajtos:
This change may introduce an endless loop if an update cannot be installed (e.g. because of filesystem permissions problems, as we experienced in the past.)

- Station starts in the tray
- We detect a new version and restart the app
  - This happens so quickly that the user cannot open the app window before we restart
- Update is not installed
- Application starts in the tray, quickly detects the new version and restarts again
- and again
- and again
- ...

Possible solution:
- Using `electron-store`, remember the last version we wanted to upgrade to
- Don't restart automatically if the "new" version is the same as the one we tried to install last time


@juliangruber WDYT?

juliangruber:
Another option: Only try to update once the app has run for one minute. This should give the user enough time to break the cycle

juliangruber:
Otherwise I agree we need to track some state to prevent this loop

bajtos:
Added [bdc462c](https://github.com/filecoin-station/desktop/pull/931/commits/bdc462c9a1c6fb17608f9d16e18750288f72f7a8) to prevent infinite loop of app restarts.

@juliangruber PTAL

juliangruber (approved):
awesome! 👏

PR #932, opened 2023-08-25
feat: report auto-updater errors to Sentry

bajtos:
At the moment, we are not notified about any auto-updater errors encountered by our users. These errors are logged and that's it.

This change adds Sentry integration to report these errors to Sentry.

Example report:
https://protocol-labs-ip.sentry.io/issues/4420725013/?project=6762462


bajtos:
The fix does not really work :(
https://protocol-labs-ip.sentry.io/issues/4420820403/?project=6762462

I guess we have to live with error reports without the stack trace. The title says _MacUpdater.onUpdaterError_, so at least we know what's the report about.

juliangruber (approved):
Thank you for looking into how to improve stack traces, is the issue on Sentry's or Electron's side? Otherwise we could try attaching err.stack manually?

juliangruber:
See for example https://github.com/getsentry/sentry-javascript/issues/1607#issuecomment-425974249

bajtos:
I am not sure. I thought the problem was caused by electron-updater giving us only a string message in `err`; that's why I tried to wrap that message into a new Error object in https://github.com/filecoin-station/desktop/pull/932/commits/f406daf152e510ca7aa6fc88f033ac9fab39c0de. Unfortunately, that did not help - Sentry showed `Error` instead of my custom `err.name` value and did not show any stack trace in the UI. (This second problem could have been caused by the fact that we didn't tell Sentry where to find the source code for the particular Station version reporting the problem, because this was reported from a local dev build.)


Iterating on this is time-consuming, as I have to wait for `npm run package` to build a new universal binary & installer.

Let's land the current version, as it brings most of the improvement, and defer further improvements until they are needed from our experience.

PR #946, opened 2023-09-04
ci: use WIN_* prefix for CSC config

bajtos:
Rename the following environment variables in the packaging step:
- `CSC_LINK` to `WIN_CSC_LINK`
- `CSC_KEY_PASSWORD` to `WIN_CSC_KEY_PASSWORD`

The goal is to fix the following error reported by our CI:

```
Env WIN_CSC_LINK is not correct, cannot resolve: D:\a\desktop\desktop not a file
```


bajtos:
See https://github.com/filecoin-station/desktop/pull/944#issuecomment-1704882268

bajtos:
The CI passed, and the problem seems fixed.

I am not sure, though, what will happen when we run a release build - will we get a properly signed binary?

juliangruber (approved):
> The CI passed, and the problem seems fixed.
> 
> I am not sure, though, what will happen when we run a release build - will we get a properly signed binary?

I don't know! According to [docs](https://www.electron.build/code-signing), there shouldn't even be a difference in the first place:

Screenshot 2023-09-04 at 12 28 12

You can next create a pre-release, and I can test it on my windows machine.

juliangruber:
Yes and yes 👍  Or use `npm version`

bajtos:
The draft release: https://github.com/filecoin-station/desktop/releases/tag/untagged-a3a59818a065198f8606

juliangruber:
The release opened without issues 👍

PR #999, opened 2023-09-29
Switch to 0x address. Closes #994

### LINKED ISSUE #994: Add 0x wallet
Either switch from the f1 to a 0x wallet, or replace it. We need the 0x address for easy integration with smart contracts.

Slack discussions:
- https://filecoinproject.slack.com/archives/C03PQG6UT2B/p1695727060317549
```[tasklist]
### Tasks
- [x] Switch to eth stack
- [x] Display f4 address
- [x] Transfer funds to 0x address
- [x] Transfer funds to f4 address
- [x] Transfer funds to f1 address (FilForwarder)
```

juliangruber:
Closes #994


juliangruber:
@bajtos do you know why socket reports removed dependencies only?

bajtos:
> do you know why socket reports removed dependencies only?
(not high priority)

I noticed that too, but don't know why it's happening.

main/wallet.js
```diff
@@ -21,7 +21,7 @@ const walletStore = new Store({
 
 const backend = new WalletBackend({
   async onTransactionUpdate () {
-    walletStore.set('transactions', backend.transactions)
+    walletStore.set('transactions_0x', backend.transactions)
```
bajtos, line 24: Can we use the same seed phrase? I think the seed phrase format standardised and it's possible to generate different addresses (BTC, ETH, FIL) from the same seed.

See:
- [BIP-0039 Mnemonic code for generating deterministic keys](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki)
- [BIP-0032 Hierarchical Deterministic Wallets](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki)
- https://github.com/satoshilabs/slips/blob/master/slip-0010.md
- https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki
- https://github.com/bitcoin/bips/blob/master/bip-0049.mediawiki
- https://github.com/bitcoin/bips/blob/master/bip-0084.mediawiki

I am concerned about the upgrade path and losing access to funds stored in existing Station wallets.

bajtos: Ouch, I got confused. This is not the seed phrase but the local storage.

Please ignore my comment above.

renderer/src/components/FilAddressForm.tsx
```diff
@@ -11,6 +15,18 @@ interface FilAddressFormProps {
   enableEditMode: () => void;
 }
 
+const checkAddressString = (address: string) => {
+  if (address.startsWith('0x')) {
+    delegatedFromEthAddress(address)
+  } else if (address.startsWith('f4')) {
+    ethAddressFromDelegated(address)
+  } else if (address.startsWith('f1')) {
+    newFromString(address)
```
bajtos, line 24: Should we warn the user that sending funds to f1 addresses is more expensive because we have to go through the FilForwarder smart contract? Or is this a technical detail that we don't want to bother our non-technical users with?

juliangruber: Yes! Great that you bring this up. I think we can show a warning, hopefully we have enough styles already. Let's display this when the address is being entered, but not afterwards?

juliangruber: Screenshot 2023-10-18 at 20 15 28

Like this?

juliangruber: Screenshot 2023-10-18 at 20 15 54

Like this?

bajtos: Nice! 🤩

renderer/src/components/FilAddressForm.tsx
```diff
@@ -92,7 +108,21 @@ const FilAddressForm: FC<FilAddressFormProps> = ({
...
           className={`
             absolute text-body-2xs text-white mt-3 ease-[cubic-bezier(0.85,0,0.15,1)] duration-500
-            ${(internalEditMode && addressIsValid) ? 'visible' : 'invisible opacity-0'}
+            ${(internalEditMode && addressIsValid && inputAddr.startsWith('f1'))
+              ? 'visible'
+              : 'invisible opacity-0'
+            }
+          `}
+        >
+          Warning: f1 addresses require extra gas for the FilForwarder contract
```
bajtos, line 117: I am proposing to clarify this warning - it's not the f1 address but the transaction sending money that requires extra gas fee.
```suggestion
          Warning: Sending rewards to f1 addresses incurs a higher gas fee. Consider using an f4 address instead.
```

@patrickwoodhead Could you suggest a better way how to convey this information? Leaning on you as the only native speaker in our team 😄

patrickwoodhead: What about "Sending rewards to an f1 address incurs a higher gas fee than sending to an f4 address because it invokes the FilForwarder contract. Consider using an f4 address instead."

juliangruber: Screenshot 2023-10-19 at 12 44 35

We're almost running out of space here. It doesn't yet cause layout issues, it isn't pretty, but it works

bajtos: LGTM.

I would add a bit of space between the warning and the header "TOTAL FIL". Or perhaps wrap the warning in a border box (`border: solid 1px`). Not a big deal, I don't want to bikeshed.

renderer/src/components/WalletModule.tsx
```diff
@@ -68,12 +69,24 @@ const WalletModule: FC<PropsWallet> = ({ isOpen = false }) => {
...
+          {stationAddress0x.slice(0, 6)}
+          {' '}. . .{' '}
+          {stationAddress0x.slice(stationAddress0x.length - 6, stationAddress0x.length)}
+        </p>
+        <button
+          className={'h-full w-auto flex flex-row items-center justify-start group ml-2 cursor-pointer'}
+          tabIndex={1}
+          onClick={() => navigator.clipboard.writeText(stationAddress0x) }
+        >
+          <CopyIcon className="btn-icon-primary mr-1 h-[12px] opacity-80 group-hover:opacity-100" />
```
juliangruber, line 89: We need to display 2 addresses now, for compatibility with different wallets:

Screenshot 2023-10-24 at 15 03 16

main/wallet-backend.js
```diff
@@ -69,119 +83,160 @@ class WalletBackend {
...
+          },
+          query: 'query Message($cid: String!) {message(cid: $cid) {cid}}'
+        }),
+        method: 'POST'
+      })
+      const body = await res.json()
+      if (body.data.message) {
+        return body.data.message.cid
+      }
+      console.log(body.errors)
```
bajtos, line 226: Do we have Sentry integration available here? Should we report the error to Sentry in such case?

main/wallet-backend.js
```diff
@@ -69,119 +83,160 @@ class WalletBackend {
...
     }
   }
 
+  /**
+   * @param {string} hash
+   * @returns {Promise<string>}
+   */
+  async convertEthTxHashToCid (hash) {
+    console.log('convertEthTxHashToCid', { hash })
+    for (let i = 0; i < 10; i++) {
```
bajtos, line 206: Let's use a more descriptive variable than `i` please, to make it more clear what is the intention of this loop. How about `attempts`?

bajtos (approved):
I am not able to comprehend all details, but I don't see any obvious problems.

Great work! 👏🏻

main/wallet-backend.js
```diff
@@ -49,18 +55,26 @@ class WalletBackend {
...
       ({ seed, isNew } = await this.getSeedPhrase())
     }
-    this.provider = new Filecoin(new HDWalletProvider(seed), {
-      apiAddress: 'https://api.node.glif.io/rpc/v0'
-    })
-    this.address = await this.getAddress()
+    this.provider = new ethers.providers.JsonRpcProvider(
+      'https://api.node.glif.io/rpc/v0'
+    )
+    // this.provider.on('debug', d => console.log(JSON.stringify(d, null, 2)))
```
patrickwoodhead, line 61: do we wanna include this in the commit as a comment?

main/wallet-backend.js
```diff
@@ -69,119 +83,160 @@ class WalletBackend {
...
-    assert(this.provider)
-    assert(this.address)
-    return await this.provider.getBalance(this.address)
+  /**
+   * @param {string} to
+   * @param {ethers.BigNumber} amount
+   * @returns {Promise<string>}
+   */
+  async transferFunds (to, amount) {
+    console.log('transferFunds()', { to, amount: amount.toString() })
```
patrickwoodhead, line 104: Do we want to make this more human readable if we are using the console to show transfers?

main/wallet-backend.js
```diff
@@ -69,119 +83,160 @@ class WalletBackend {
...
-      gasPremium: 0,
-      gasFeeCap: 0,
-      gasLimit: 0
+  async transferFundsToF1Address (to, amount) {
+    return await this.runTransaction(to, amount, async () => {
+      assert(this.signer)
+      assert(this.filForwarder)
+
+      const amountMinusGas = amount.sub(ethers.utils.parseEther('0.1'))
+      console.log('filForwarder.forward()', {
```
patrickwoodhead, line 128: same question here

main/wallet-backend.js
```diff
@@ -69,119 +83,160 @@ class WalletBackend {
...
+   * @param {string} to
+   * @param {ethers.BigNumber} amount
+   * @returns {Promise<string>}
+   */
+  async transferFundsToEthAddress (to, amount) {
+    return await this.runTransaction(to, amount, async () => {
+      assert(this.signer)
+
+      const amountMinusGas = amount.sub(ethers.utils.parseEther('0.01'))
+      console.log('sendTransaction()', {
```
patrickwoodhead, line 149: and here

main/wallet-backend.js
```diff
@@ -69,119 +83,160 @@ class WalletBackend {
...
-        message.toLotusType()
-      )
-      const lotusMessage = messageWithGas.toLotusType()
-      const msgValid = await this.provider.simulateMessage(lotusMessage)
-      assert(msgValid, 'Message is invalid')
-      const signedMessage = await this.provider.wallet.sign(from, lotusMessage)
-      const { '/': cid } = await this.provider.sendMessage(signedMessage)
-
+      const tx = await fn()
+      console.log({ tx })
```
patrickwoodhead, line 185: again here

main/wallet-backend.js
```diff
@@ -69,119 +83,160 @@ class WalletBackend {
...
+      assert(tx.hash, 'Transaction hash not found')
+      const cid = await this.convertEthTxHashToCid(tx.hash)
+      console.log({ cid })
```
patrickwoodhead, line 188: again here

renderer/src/components/WalletModule.tsx
patrickwoodhead: I wonder if we need a ? icon that opens a modal describing why there are two addresses?

main/wallet-backend.js
bajtos: Maybe we can use `debug` or `electron-log` instead of `console.log` and that way keep this line always in & uncommented?

main/wallet-backend.js
bajtos: Actually, I think we shouldn't be using `console.log`, I think that does not end up being saved in log files and therefore these messages are lost when running the app in the tray. We should use `electron-log` instead.

renderer/src/components/WalletModule.tsx
juliangruber: This would be an improvement for sure. I don't think however we have a UI element for this :| Would you consider this a release blocker? I think we can get away without

main/wallet-backend.js
juliangruber: Errors here are expected, when the transaction can't yet be found. These aren't critical, only debug errors

main/wallet-backend.js
juliangruber: good idea, used `attempt`

main/wallet-backend.js
juliangruber: This is too noisy to keep in, I've removed it for now

main/wallet-backend.js
juliangruber: Good idea, integrated `electron-log`. @patrickwoodhead only devs see these, so JSON formatting is ok

renderer/src/components/WalletModule.tsx
juliangruber: Seeing that you approved the PR, I think this isn't a release blocker

main/wallet-backend.js
juliangruber: I've changed to only log if we see an error here that we're not expecting

renderer/src/components/FilAddressForm.tsx
juliangruber: I assume we'll have to revisit this anyway, I'd rather get this out quicker

juliangruber: And gain operational experience

bajtos (commented):
👏🏻

PR #1021, opened 2023-10-11
Fix empty wallet address during keychain permission check

### LINKED ISSUE #609: Handle keychain access delays and errors
When Filecoin Desktop starts on macOS, the user is prompted to allow access to `filecoin-station-wallet" stored in the keychain.

Screenshot 2023-03-29 at 11 49 48

## 1. Error path

The user denies access. Station reports an unknown error.

Screenshot 2023-03-29 at 11 50 07

Station starts, the tray icon reports offline mode.

Activity log mentions this error too.

Screenshot 2023-03-29 at 11 51 26

**I find this confusing from the perspective of a non-technical user. What is the error trying to tell me and how can I fix the problem?**

## 2. Happy path with delay

Restarting the Station, granting it a one-time access to the item stored in my keychain. Station happily starts now.

When I open the Wallet UI, I don't see any wallet address 😱

Screenshot 2023-03-29 at 11 53 39

The button "Copy" places an empty string to the clipboard.

Opening and closing the window does not fix the issue. (Should we start closing the browser window instead of hiding it?)

Restarting the Station does not help either, because the keychain prompt is shown again.

The only solution I found is to force reload of the UI using the shortcut I know from my browser: <kbd>Command</kbd>+<kbd>R</kbd>

juliangruber:
Fixes #609

Alternative implementation to https://github.com/filecoin-station/desktop/pull/658, which this is based on. Thank you, @taesuklee!

Closes #658

### DIFF
--- a/main/core.js
+++ b/main/core.js
@@ -50,13 +50,13 @@ async function setup (ctx) {
  * @param {Context} ctx
  */
 async function start (ctx) {
-  assert(wallet.getAddress(), 'Core requires FIL address')
+  assert(await wallet.getAddress(), 'Core requires FIL address')
   console.log('Starting Core...')
 
   const childProcess = fork(corePath, ['--json'], {
     env: {
       ...process.env,
-      FIL_WALLET_ADDRESS: wallet.getAddress(),
+      FIL_WALLET_ADDRESS: await wallet.getAddress(),
       CACHE_ROOT: consts.CACHE_ROOT,
       STATE_ROOT: consts.STATE_ROOT,
       DEPLOYMENT_TYPE: 'station-desktop'
--- a/main/core.js
+++ b/main/core.js
@@ -50,7 +50,6 @@ async function setup (ctx) {
  * @param {Context} ctx
  */
 async function start (ctx) {
-  assert(await wallet.getAddress(), 'Core requires FIL address')
   console.log('Starting Core...')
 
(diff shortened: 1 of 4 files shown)


main/core.js
```diff
@@ -50,13 +50,13 @@ async function setup (ctx) {
...
 async function start (ctx) {
-  assert(wallet.getAddress(), 'Core requires FIL address')
+  assert(await wallet.getAddress(), 'Core requires FIL address')
   console.log('Starting Core...')
 
   const childProcess = fork(corePath, ['--json'], {
     env: {
       ...process.env,
-      FIL_WALLET_ADDRESS: wallet.getAddress(),
+      FIL_WALLET_ADDRESS: await wallet.getAddress(),
```
bajtos, line 59: Do we have to await `getAddress` call twice? Can that delay startup? Have you considered storing the address in a variable instead?

```js
async function start (ctx) {
  const walletAddress = await wallet.getAddress()
  assert(walletAddress, 'Core requires FIL address')
  // ...
     FIL_WALLET_ADDRESS: walletAddress
```

bajtos (approved):

main/core.js
juliangruber: good point!

bajtos (approved):
:shipit: 

_(I think you need to merge in `main`.)_

PR #1035, opened 2023-10-17
Wallet: fix dates. Closes #1034

### LINKED ISSUE #1034: Station wallet: The wallet history dates are broken
When I open my wallet, the date of the only transaction is 20/01/1970.

image

Station Version: 0.24.1
OS: MacOS Ventura 13.5

juliangruber:
Closes #1034

filfox changed their response format

### DIFF
--- a/main/test/wallet-backend.test.js
+++ b/main/test/wallet-backend.test.js
@@ -47,8 +47,19 @@ describe('Wallet Backend', function () {
     it('fetches all transactions', /** @this {Mocha.Test} */ async function () {
       this.timeout(20_000)
 
+      await backend.setup(
+        // eslint-disable-next-line max-len
+        'insane believe defy best among myself mistake account paddle episode life music fame impact below define habit rotate clay innocent history depart slice series'
+      )
       await backend.fetchAllTransactions()
-      assert(backend.transactions)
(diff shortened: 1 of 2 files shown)


bajtos (approved):
Have you considered adding a test for this?

1. Setup the wallet backend with a known address with some transactions
2. Call `fetchAllTransactions`
3. Find a known transaction in the list of all txs
4. Assert that the transaction has expected fields & values

juliangruber:
Good idea! Can you please take another look?

main/test/wallet-backend.test.js
```diff
@@ -47,8 +47,19 @@ describe('Wallet Backend', function () {
     it('fetches all transactions', /** @this {Mocha.Test} */ async function () {
       this.timeout(20_000)
 
+      await backend.setup(
+        // eslint-disable-next-line max-len
+        'insane believe defy best among myself mistake account paddle episode life music fame impact below define habit rotate clay innocent history depart slice series'
```
bajtos, line 52: It's probably obvious, but I want to call it out explicitly:

All wallets derived from this seed are public, meaning that anyone can sign transactions using wallet's private key.

bajtos (approved):
NOICE 👏🏻

PR #1075, opened 2023-11-05
Add reject sanctioned addresses. Closes #1074

### LINKED ISSUE #1074: Wallet screening
Screen destination wallet addresses, for legal reasons

- [x] Create API `GET /valid/:address` -> `200 OK` or `403 Forbidden` https://github.com/filecoin-station/wallet-screening
- [x] Query forbidden addresses
- [ ] Add async validation to destination address input field https://github.com/filecoin-station/desktop/pull/1075

juliangruber:
Closes #1074

### DIFF
--- a/renderer/src/components/FilAddressForm.tsx
+++ b/renderer/src/components/FilAddressForm.tsx
@@ -15,7 +15,7 @@ interface FilAddressFormProps {
   enableEditMode: () => void;
 }
 
-const checkAddressString = (address: string) => {
+const checkAddressString = async (address: string) => {
   if (address.startsWith('0x')) {
     delegatedFromEthAddress(address)
   } else if (address.startsWith('f4')) {
@@ -24,6 +24,10 @@ const checkAddressString = (address: string) => {
     newFromString(address)
   } else {
     throw new Error('Invalid address type')
+  }
+  const res = await fetch(`https://station-wallet-screening.fly.dev/${address}`)
+  if (res.status === 403) {
+    throw new Error('Sanctioned address')
   }
 }
 
@@ -58,12 +62,14 @@ const FilAddressForm: FC<FilAddressFormProps> = ({
     if (inputAddr === '') {
       setAddressIsValid(true)
     } else {
-      try {
-        checkAddressString(inputAddr)
-        setAddressIsValid(true)
-      } catch {
-        setAddressIsValid(false)
-      }
+      (async () => {
+        try {
+          await checkAddressString(inputAddr)
+          setAddressIsValid(true)
+        } catch {
+          setAddressIsValid(false)
+        }
+      })()
     }
   }, [inputAddr])
 


bajtos:
@juliangruber Can you please share an example of a sanctioned wallet address that I can use to test this new feature?

renderer/src/components/FilAddressForm.tsx
```diff
@@ -25,6 +25,10 @@ const checkAddressString = (address: string) => {
   } else {
     throw new Error('Invalid address type')
   }
+  const res = await fetch(`https://station-wallet-screening.fly.dev/${address}`)
```
bajtos, line 28: Have you considered moving this check to the `main` part?

I see the following benefits:
- We don't need the screening service to support CORS
- It may be easier to write tests for this check

juliangruber:
> @juliangruber Can you please share an example of a sanctioned wallet address that I can use to test this new feature?

Sorry, yes: 0x1da5821544e25c636c1417ba96ade4cf6d2f9b5a

renderer/src/components/FilAddressForm.tsx
juliangruber: I thought about it, yes!

> * We don't need the screening service to support CORS

Adding CORS support to the screening service is a lot less logic than adding another Desktop IPC call unfortunately. Plus, debugging is easier when you see the HTTP request in the network tab.

> * It may be easier to write tests for this check

I propose to reevaluate that when we are going to write the test

bajtos:
The new check seems to work, but I find the user interaction confusing. The error message says _The FIL address entered is invalid. Please check and try again._ But the address `0x1da5821544e25c636c1417ba96ade4cf6d2f9b5a` is a valid one and trying to enter it again (as suggested by the app) is not going for solve the issue for me.

Can we show a more helpful error to the user please?
- Explain that we cannot send funds to that wallet for legal reasons.
- Ask them to use a different wallet address instead.

renderer/src/components/FilAddressForm.tsx
bajtos: Fair enough 👍🏻

juliangruber:
I've decided initially not to create a different address because
- how important is it? how many people will have sanctioned wallets? I assume with our user base and existing geo fencing, close to none
- we can even assume, that if a sanctioned wallet is used, people are trying to bypass the system, in which case Station doesn't need to be helpful

Do you still think we should improve that?

juliangruber:
> Explain that we cannot send funds to that wallet for legal reasons.

This is part of onboarding, from that perspective a sanctioned valid is an invalid address

bajtos (approved):
I see your points, I agree to not spend too much effort to handle the situation that should be extremely rare.

juliangruber:
I will fix CI in a separate PR (unrelated to this change)

PR #1093, opened 2023-11-09
test: change wallet for testing scheduled rewards

bajtos:
The seed used for testing `fetchAllTransactions` is used in other places, too and as a result, the address has some rewards accrued in the contract.

That breaks the test for `fetchScheduledRewards`.

This commit fixes the problem by generating a new random wallet seed for this test.


bajtos:
@juliangruber hindsight please 🙏🏻

juliangruber:
Perfect!

PR #1097, opened 2023-11-10
Fix only show update when it is downloaded. Closes #137

### LINKED ISSUE #137: Show auto-updater status in the top banner when an update is pending
See the Sigma prototype:

Screenshot_2022-09-19_at_15 45 33

See the [Electron docs](https://www.electron.build/auto-update#events) for the full list of events emitted by the auto-updater module. I think we may want to visualise the following:
- `update-available`
- `download-progress`
- `update-downloaded`

juliangruber:
Closes #137

Instead of showing different update states for when an update is available but not yet downloaded, download progress, and the update is available, this just changes it so that an update notification is shown once an update is downloaded and available. I don't think users care about in-between events, just want to know when they can restart to update.

### DIFF
--- a/main/updater.js
+++ b/main/updater.js
@@ -134,28 +134,11 @@ function onUpdateAvailable ({ version /*, releaseNotes */ }) {
   updateAvailable = true
   nextVersion = version
 
-  ipcMain.emit(ipcMainEvents.UPDATE_AVAILABLE)
   log.info(`Update to version ${version} is available, downloading..`)
   autoUpdater.downloadUpdate().then(
     _ => log.info('Update downloaded'),
     err => log.error('Cannot download the update.', err)
   )
-
-  if (!checkingManually) { return }
-  // do not toggle checkingManually off here so we can show a dialog once the
-  // download is finished.
-
-  const buttonIx = showDialogSync({
-    title: 'Update available',
-    message: `A new version ${version} of Filecoin Station is available. ` +
-      'The download will begin shortly in the background.',
-    type: 'info',
-    buttons: ['Close', 'Show Release Notes']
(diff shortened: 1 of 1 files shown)


main/updater.js
```diff
@@ -134,28 +134,11 @@ function onUpdateAvailable ({ version /*, releaseNotes */ }) {
...
-    title: 'Update available',
-    message: `A new version ${version} of Filecoin Station is available. ` +
-      'The download will begin shortly in the background.',
-    type: 'info',
-    buttons: ['Close', 'Show Release Notes']
-  })
-
-  if (buttonIx === 1) {
-    openReleaseNotes()
-  }
```
bajtos, line 158: IIUC, after your changes, when I trigger a manual check, I will no longer have an easy way to find the release notes for the new version.


The dialog opened by `onUpdateDownloaded` has only these two buttons:
```
 buttons: ['Later', 'Install now']
```

I propose to add the button "Show Release Notes" to `onUpdateDownloaded` dialog.

WDYT?

bajtos (commented):
Great idea! I agree users don't care about the event "a new version is available but not downloaded yet".

main/updater.js
juliangruber: You're right, release notes are now only linked to from the update banner. Added!

bajtos (approved):
I don't have the bandwidth to test this change locally myself.

One thing to consider: when the app is open, and a new version is detected, we show a banner at the top of the window. It would be good to verify that this works after your changes.

The code changes LGTM, I don't see any obvious problems.

juliangruber:
Verified and works 👍

PR #1168, opened 2023-12-04
Read scheduled rewards from `core`

patrickwoodhead:
The desktop no longer makes a request to the contract to fetch the scheduled rewards. Instead it gets this info from core.

The balance sent around by the IPC events is now not truncated to 6 decimal places but instead it is the full 18 significant figures of the balance. This leaves it to each of the renderers (UI, tray etc) to decide how they want to format the value.


bajtos:
@patrickwoodhead thank you for the pull request! Can you please use a more descriptive PR title than "1134 refactor"? For example: "refactor scheduled rewards"

main/wallet.js
```diff
@@ -147,9 +131,14 @@ async function updateScheduledRewards () {
 
 async function _updateScheduledRewards () {
   assert(ctx)
-  scheduledRewards = await backend.fetchScheduledRewards()
```
juliangruber, line 150: if we don't update `scheduledRewards`, future calls to `getScheduledRewards` will return the outdated value

renderer/src/utils/number-ops.ts
```diff
@@ -0,0 +1,5 @@
+export function roundToSixDecimalPlaces (number: string | number | undefined) {
+  if (!number) return 0
+  const roundedNumber = Number(number).toFixed(6) // Round to 6 decimal places
+  return roundedNumber.toString()
+}
```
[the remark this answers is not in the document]
juliangruber: @bajtos I think it's a module system issue: Client side code crashed on `module.exports`, electron side code crashed on `exports`.

renderer/src/utils/number-ops.ts
```diff
@@ -0,0 +1,7 @@
+export function formatTokenValue (input: string | number | undefined): number {
```
juliangruber, line 1:
```suggestion
// Keep in sync with main/utils.js
export function formatTokenValue (input: string | number | undefined): number {
```

main/utils.js
```diff
@@ -1,12 +1,17 @@
 'use strict'
 
 /**
- * @param { number | string | undefined } amount
- * @returns { string }
- */
+   * @param {string | number | undefined} input
+   * @returns {number}
+*/
+function formatTokenValue (input) {
```
juliangruber, line 7:
```suggestion
// Keep in sync with renderer/src/number-ops.ts
function formatTokenValue (input) {
```

juliangruber (changes requested):

main/wallet.js
patrickwoodhead: We update it in `main/core.js` line 88 with `wallet.setScheduledRewards`

patrickwoodhead: i.e. it is updated from core rather than from the backend

juliangruber: Yes but the value in the backend will be wrong afterwards, right?

patrickwoodhead: Unless I am wrong, we don't use the value in the backend anywhere, or store it as state in the backend. Global search for `fetchScheduledRewards` just shows the function and its test now so could even delete it.

juliangruber: Yeah, but someone could start using it again. If you want to keep it static then please add a code comment that it will get outdated

renderer/src/utils/number-ops.ts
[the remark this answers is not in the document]
bajtos: Oh, I forgot we use CommonJS in the main and ESM in the renderer.

I think there are ways to get around that, but I agree it's not necessary for this pull request.

main/wallet.js
patrickwoodhead: Ok good catch thanks - I now make sure to fetch from the wallet store each time, not from the state in the wallet file.

juliangruber (approved):
NICE!

bajtos:
@juliangruber @patrickwoodhead, what's the status of this PR? Are there any changes needed before we can land & release this?

PR #1219, opened 2024-01-02
Add auth to glif requests

juliangruber:
Ref https://filecoinproject.slack.com/archives/C023K7D9GAX/p1703679868805479

### DIFF
--- a/main/wallet-backend.js
+++ b/main/wallet-backend.js
@@ -56,9 +56,12 @@ class WalletBackend {
     } else {
       ({ seed, isNew } = await this.getSeedPhrase())
     }
-    this.provider = new ethers.providers.JsonRpcProvider(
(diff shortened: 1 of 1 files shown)


bajtos (approved):
TL;DR: I am okay with landing this fix as a stop-gap solution, but I would like us to consider the ramifications.

It's great that you are using unique access tokens per repository 👏🏻 

Longer comment: https://github.com/filecoin-station/core/pull/305#pullrequestreview-1800713231

Wallet-backend has full access to Station's built-in wallet, can we use that wallet to generate a new access token at startup? (If that's the way how access tokens are created.)

juliangruber:
See https://github.com/filecoin-station/core/pull/305#issuecomment-1875158080

PR #1296, opened 2024-02-02
`filecoin-station/core@18`

juliangruber:


main/activities.js
```diff
@@ -24,14 +27,19 @@ class Activities {
...
   #detectChangeInOnlineStatus (activity) {
-    if (
-      activity.type === 'info' &&
-      (activity.message === 'SPARK started reporting retrievals' ||
-      activity.message === 'SPARK retrieval reporting resumed')
-    ) {
-      this.#online = true
-    } else if (activity.message === 'SPARK failed reporting retrieval') {
-      this.#online = false
+    const firstWord = activity.message.split(' ')[0].toLowerCase()
```
juliangruber, line 30: I chose to make this a bit more dynamic, as I think it was getting too verbose and too easy to break when iterating on messages. It can produce false results too, so it's not a finished solution, but in my mind an improvement. Happy to follow up on this

juliangruber:
@bajtos hindsight please 🙏

PR #333, opened 2022-12-12
feat: wallet api - part 1/2

jopedroliveira:
### Part 1 of 2 - APIs and Structure

This PR briefly implements the required RPC calls and handlers for the Station Wallet Module.
The changes on this PR may lead to application malfunction since this is part 1 of the feature, separated here for readability purposes.

🚨 Backend: FilAddress configuration keys need refactoring. Left some // todo comments

<del>⚠️ Should be merged after #332</del>

PR #1300, opened 2024-02-05
Fix await `ui.loadURL()`. Closes #333

juliangruber:
Closes #333 

It looks like we mustn't perform any other actions until `loadURL` is finished.

Ref https://github.com/electron/electron/issues/18857#issuecomment-503012296
Ref https://github.com/electron/electron/issues/37123


juliangruber:
@bajtos hindsight please

bajtos (commented):
👏🏻

PR #1324, opened 2024-02-09
refactor: rename "logs" to "logLines"

bajtos:
Make it clear in code storing logs that chunks must be split at newlines.

My intention is to prevent regressions like the one fixed by #1319.

### DIFF
--- a/main/core.js
+++ b/main/core.js
@@ -10,7 +10,7 @@ const Sentry = require('@sentry/node')
 const consts = require('./consts')
 const { randomUUID } = require('node:crypto')
 const { Activities } = require('./activities')
-const { Logs } = require('./logs')
+const { LogLines } = require('./logs')
 const split2 = require('split2')
 const { parseEther } = require('ethers/lib/utils')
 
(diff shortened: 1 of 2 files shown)


juliangruber (changes requested):
I don't think this is the way to go. While I agree the needs a refactor to prevent regressions, I think the fault lies inside the abstraction. To me, the library is still for logs (it's an implementation detail whether it's storing individual lines or not). Then, I see too ways to make it safer:

a) Change the method names to `Logs#pushLine()` and `Logs#getLastLines()`
b) Inside `Logs#push()`, split strings on newlines, so that whatever can be pushed to it

What do you think?

bajtos:
> I don't think this is the way to go. While I agree the needs a refactor to prevent regressions, I think the fault lies inside the abstraction. To me, the library is still for logs (it's an implementation detail whether it's storing individual lines or not). 

👍🏻 

> Then, I see too ways to make it safer:
> 
> a) Change the method names to `Logs#pushLine()` and `Logs#getLastLines()` 

I did consider this option, too, and I am happy to adopt it.

> b) Inside `Logs#push()`, split strings on newlines, so that whatever can be pushed to it

This goes beyond a simple cleanup, it will require a bit more work & testing, which I don't have appetite for right now.

juliangruber:
> I did consider this option, too, and I am happy to adopt it.

Go for it! 🙏

bajtos:
Closing in favour of https://github.com/filecoin-station/desktop/pull/1325

PR #1325, opened 2024-02-13
refactor: rename `logs.push()` to `logs.pushLine()`

bajtos:
Make it clear in code storing logs that chunks must be split at newlines.

My intention is to prevent regressions like the one fixed by #1319.

### DIFF
--- a/main/logs.js
+++ b/main/logs.js
@@ -13,7 +13,7 @@ class Logs {
     this.#logs.splice(0, this.#logs.length - 100)
   }
 
-  getLines () {
+  get () {
     return this.getLastLines(Infinity)
   }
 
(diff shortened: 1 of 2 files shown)


main/logs.js
```diff
@@ -8,20 +8,20 @@ class Logs {
...
    */
-  push (line) {
+  pushLine (line) {
     this.#logs.push(line)
     this.#logs.splice(0, this.#logs.length - 100)
   }
 
-  get () {
-    return this.getLast(Infinity)
+  getLines () {
```
juliangruber, line 16:
```suggestion
  get () {
```

or

```suggestion
  getAll () {
```

Since this is returning all, I find the `Lines` part confusing. But I don't feel strongly about this, please ignore if you disagree.

juliangruber (approved):
👏

main/logs.js
bajtos: We are keeping only the last 100 lines (IIRC). I feel "getAll" would be misleading.

I don't mind to use `get`.

PR #1337, opened 2024-02-20
fix missing `await` for `loadURL`. for #333

juliangruber:
Follow up for https://github.com/filecoin-station/desktop/pull/1300, found another missing `await`.


bajtos (approved):
👏🏻 

There is a TypeScript-aware ESlint rule that can catch these problems, but IIRC, these rules don't work for JS files using JSDoc comments for types. 😢

PR #1345, opened 2024-02-21
updater: fix update is only available after download

juliangruber:


main/updater.js
```diff
@@ -54,7 +54,7 @@ module.exports = async function setupUpdater (
   /** @type {import('./typings').Context} */ ctx
 ) {
   ctx.getUpdaterStatus = function getUpdaterStatus () {
-    return { updateAvailable }
+    return { updateAvailable: updateDownloaded }
```
bajtos, line 57: I would prefer to rename the places reading this property too. Otherwise, we are using different names ("update available" vs "update downloaded") for the same thing and this could create confusion in the future.

bajtos (approved):

main/updater.js
juliangruber: I was thinking about this too, but decided against this because I think the places outside this library shouldn't have to care about the finer details of updating. I could see however how `updateDownloaded` is more clear. I was just thinking that as an outsider, "update downloaded" could be more confusing than "update available" (ok it's downloaded, are there any other steps?). Maybe instead something like `readyToUpdate`?

juliangruber: what do you think about this?

bajtos: I agree with your reasoning. 👍🏻 

I like the idea of renaming the property to `readyToUpdate` or perhaps `updateReady`.

bajtos (approved):
Oh wow, that's a really involved change. It would be better to land the last commit in a standalone PR. Otherwise, it's not clear which changes are related to refactor-rename and which changes are modifying the behaviour.

I skimmed through the changes and didn't see any obvious issues. I did not test the new version manually, though.

juliangruber:
https://github.com/filecoin-station/desktop/pull/1353

PR #1383, opened 2024-03-05
wallet: only update transactions when UI visible

juliangruber:
Ref https://github.com/filecoin-station/desktop/issues/1378


bajtos (approved):
Nice! A bit hacky but good enough to fix the issue quickly.

PR #1389, opened 2024-03-07
updater: add retry `downloadUpdate()`

juliangruber:
Ref https://filecoinproject.slack.com/archives/C03PVLW721L/p1709739129656009

### DIFF
--- a/main/updater.js
+++ b/main/updater.js
@@ -10,6 +10,9 @@ const log = require('electron-log').scope('updater')
 const { showDialogSync } = require('./dialog')
 const Store = require('electron-store')
 
+/** @type {import('p-retry').default} */
+let pRetry
+
 // must be global to avoid gc
 let updateNotification = null
 
@@ -34,7 +37,9 @@ function beforeQuitCleanup () {
   app.removeAllListeners('window-all-closed')
 }
 
-function setup (/** @type {import('./typings').Context} */ ctx) {
+async function setup (/** @type {import('./typings').Context} */ ctx) {
+  pRetry = (await import('p-retry')).default
+
   autoUpdater.logger = log
   autoUpdater.autoDownload = false // we download manually in 'update-available'
 
@@ -134,7 +139,16 @@ function onUpdateAvailable ({ version /*, releaseNotes */ }) {
   nextVersion = version
 
   log.info(`Update to version ${version} is available, downloading..`)
-  autoUpdater.downloadUpdate().then(
+  pRetry(
+    () => autoUpdater.downloadUpdate(),
+    {
+      retries: 10,
+      onFailedAttempt: err => {
+        console.error(err)
+        console.error('Failed to download update. Retrying...')
+      }
+    }
+  ).then(
     _ => log.info('Update downloaded'),
     err => log.error('Cannot download the update.', err)
   )


main/updater.js
```diff
@@ -34,7 +37,9 @@ function beforeQuitCleanup () {
   app.removeAllListeners('window-all-closed')
 }
 
-function setup (/** @type {import('./typings').Context} */ ctx) {
+async function setup (/** @type {import('./typings').Context} */ ctx) {
```
bajtos, line 40: If we change setup to an async function, then I think we should also modify the code calling this function to await the outcome, right?

How about moving the `await` part into `onUpdateAvailable`?

```js
// at the top
const pRetryPromise = import ('p-retry').then(({default}) => (default))

// in `onUpdateAvailable`
pRetryPromise
.then(pRetry => pRetry(
   () => autoUpdater.downloadUpdate(),
   // etc.
)
.then(
    _ => log.info('Update downloaded'),
    err => log.error('Cannot download the update.', err)
  )
```

bajtos (commented):
I think this is a good improvement, but I don't think it's going to fix all edge cases. If I am on a long train ride with a spotty internet connection, all 10 retries may happen while I am offline. Then, I end up in the same broken state.

Given that, I am not sure if it's worth landing this PR at all.

WDYT?

A more robust solution to consider:
- Change the "check for update" frequency from 12 hours to 1 hour
- When checking for a new version, check our state. If there is a new version that we could not download before, then trigger the download. If there is no new version, then trigger the check for a new version.

Alternatively, we can also enable the electron-updater's auto-download feature; maybe it will work better. It makes me wonder why we have disabled it in the past.
-

main/updater.js
juliangruber: This function is called in https://github.com/filecoin-station/desktop/blob/866e0887e4548189004fcb099bc3ef040f3a63d8/main/index.js#L173, where it's already being awaited. Or am I missing something?

bajtos:
> > I think this is a good improvement, but I don't think it's going to fix all edge cases. If I am on a long train ride with a spotty internet connection, all 10 retries may happen while I am offline. Then, I end up in the same broken state.
> > Given that, I am not sure if it's worth landing this PR at all.
> 
> I agree this doesn't solve all cases, but it can handle more cases than the current implementation. Plus, it doesn't introduce any complexity in how we're handling our state.

Fair enough 👍🏻 

> > A more robust solution to consider:
> > Change the "check for update" frequency from 12 hours to 1 hour
> > When checking for a new version, check our state. If there is a new version that we could not download before, then trigger the download. If there is no new version, then trigger the check for a new version.
> 
> Isn't this effectively the same thing as in this PR, only without exponential backoff and with infinite retries? What about instead configuring the retry logic to have 100 retries? I like this way because we're not introducing any existing state to the updater library.

This PR does introduce a new state to the updater library and this state is hidden inside pRetry logic. We have no control of that state - for example, it's not possible to abort pRetry loop trying to download version X after we detect that version X+1 is available. We can end up with multiple pRetry loops running at the same time. (Fortunately, the API for downloading the update is not version specific, so in the edge case I described, both pRetry loops will try to download the version X+1.) A possible fix is to setup an AbortController, pass the `signal` to pRetry (assuming it supports that?) and abort the retry loop when we detect a newer version. 

Personally, I prefer to be explicit and manage the state ourselves.

> > Alternatively, we can also enable the electron-updater's auto-download feature; maybe it will work better. It makes me wonder why we have disabled it in the past.
> 
> Do you have a link to this feature? I'm not finding it in their docs

Docs (very brief, unfortunately):
https://github.com/electron-userland/electron-builder/blob/9bcede88f2083d41266e48dfa712adc2d223bd7f/docs/auto-update.md#appupdater--moduleeventseventemitter

Our usage:
https://github.com/filecoin-station/desktop/blob/3c9bf056da8a3f51c97d9dd9decbb419a0a46592/main/updater.js#L39

main/updater.js
bajtos: Ah, that's great to know. Thanks for checking. All is good; my comment can be ignored.

bajtos (approved):

juliangruber:
> This PR does introduce a new state to the updater library and this state is hidden inside pRetry logic.

Agreed, I should have been more explicit here. This PR doesn't introduce any state that we need to manage. To the outside it's indistinguishable if we have one download call that takes 10 minutes, or whether we are on the 5th retry of faster performing download calls.

> We have no control of that state - for example, it's not possible to abort pRetry loop trying to download version X after we detect that version X+1 is available. We can end up with multiple pRetry loops running at the same time. (Fortunately, the API for downloading the update is not version specific, so in the edge case I described, both pRetry loops will try to download the version X+1.) A possible fix is to setup an AbortController, pass the signal to pRetry (assuming it supports that?) and abort the retry loop when we detect a newer version.

As you suggested, this is possible: https://github.com/sindresorhus/p-retry?tab=readme-ov-file#signal. To me this is also very similar to not having retry logic, but one small running download call (downloading one electron app over train internet can also take long. Admittedly not 24h, but just as long maybe as 10 retries on a faster internet.

> Personally, I prefer to be explicit and manage the state ourselves.

I disagree here, as this complicates our updater. What about we merge this as is, as it's already an improvement, and then discuss further improving this logic in a separate PR? I personally am not interested in following up at this moment, until we get new bug reports after this logic has been merged and proved to still be insufficient.

bajtos:
> What about we merge this as is, as it's already an improvement

Agreed 👍🏻

PR #1496, opened 2024-04-26
fix: formatting of startup logs

bajtos:
I noticed that electron-log file transport does not apply `%s` formatting directives.

Example output before my change:

```
[2024-04-04 10:52:22.398] [info]  (main)           Filecoin Station build version: %s %s-%s%s%s 1.4.5 darwin arm64
[2024-04-04 10:52:22.398] [info]  (main)           Machine spec: %s version %s Darwin 23.4.0
[2024-04-04 10:52:22.761] [info]  (wallet)         Using existing seed phrase
[2024-04-04 10:52:22.761] [info]  (wallet)         Address: %s 0x00...00
```

In this commit, I am adding an `util.format()` call to build the full log message string before passing it to electron-log methods.


juliangruber (approved):
👏

PR #1498, opened 2024-04-26
Connect settings page

davelange:
Add functionality to settings page:
 - Adapt current settings that are in the top menu tray
 - Add settings actions to ipcMain and preload
 - Use settings actions in the Settings page

### DIFF
--- a/main/core.js
+++ b/main/core.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { app, dialog, clipboard } = require('electron')
+const { app, dialog } = require('electron')
 const { join, dirname } = require('node:path')
 const { fork } = require('node:child_process')
 const wallet = require('./wallet')
@@ -13,7 +13,6 @@ const { Activities } = require('./activities')
 const { Logs } = require('./logs')
 const split2 = require('split2')
 const { parseEther } = require('ethers/lib/utils')
-const { showDialogSync } = require('./dialog')
 
 /** @typedef {import('./typings').Context} Context */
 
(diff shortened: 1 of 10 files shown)


main/core.js
```diff
@@ -42,6 +43,29 @@ async function setup (ctx) {
       await fs.writeFile(filePath, logs.get())
     }
   }
+
+  ctx.toggleOpenAtLogin = () => {
```
juliangruber, line 47: The `main/core.js` file is about managing the Station Core instance, not the "core" app functions. What do you think about creating `main/settings.js`, and managing these `ctx` methods there? It would also have a public `setup` method, which would be run in `main/index.js`

renderer/src/pages/settings/Settings.tsx
```diff
@@ -1,18 +1,30 @@
...
+  exportSeedPhrase,
+  isOpenAtLogin,
+  saveModuleLogsAs,
+  toggleOpenAtLogin
+} from 'src/lib/station-config'
+
 const Settings = () => {
+  async function getUpdatedCheckedValue (node: HTMLInputElement) {
+    if (node) node.checked = await isOpenAtLogin()
+  }
```
juliangruber, line 12: Can we please `useState` together with `useEffect` here, instead of manually updating the DOM node? I would like to follow React conventions as closely as possible

juliangruber (changes requested):
other than the two items, looks great!

renderer/src/pages/settings/Settings.tsx
```diff
@@ -1,18 +1,40 @@
...
+
 const Settings = () => {
+  const [isOpenAtLoginChecked, setIsOpenAtLoginChecked] = useState(true)
+
+  const updateIsOpenAtLogin = async () => setIsOpenAtLoginChecked(await isOpenAtLogin())
+
+  async function handleClick () {
+    toggleOpenAtLogin()
+    updateIsOpenAtLogin()
+  }
```
juliangruber, line 18:
```suggestion
  async function handleClick () {
    toggleOpenAtLogin()
    setIsOpenAtLoginChecked(!isOpenAtLoginChecked)
  }
```

should be sufficient

juliangruber (changes requested):

renderer/src/pages/settings/Settings.tsx
davelange: That would toggle the checkbox even if for some reason the operation fails. Should we consider that possibility? If it's not an issue, I'll make the change

juliangruber: This has never failed yet (I believe), the current version in tray.js also assumes it has worked

davelange: Ok, commited 👍

juliangruber (approved):

PR #1510, opened 2024-05-02
Update tests

davelange:
Update UI tests
- Update vitest and @vitest/ui versions
- The version update caused some tests to break, so update those tests. Mostly this means refactoring how functions are mocked.
- Add tests for the settings page.


juliangruber (approved):
😍

PR #1514, opened 2024-05-06
Add wallet modal

davelange:
Add unstyled wallet info modal:
 - Create a DialogProvider. We will need to show other modals eventually, so this provider renders the base Dialog component and receives a component to render inside that Dialog. It can be opened by calling `openDialog({content: <SomeContent />})`. Use `@radix-ui/react-dialog` for the base component, as it provides better accessibility (focus trapping, shortcuts to close, close on click outside)
 - Add the wallet modal content.


renderer/src/lib/utils.ts
```diff
@@ -0,0 +1,16 @@
...
+  return `${value.slice(0, size)}...${value.slice(-size)}`
+}
+
+export function formatFilValue (value = '') {
+  return new FilecoinNumber(String(value), 'fil')
+    .decimalPlaces(3, BigNumber.ROUND_DOWN)
+    .toString()
+}
+
+export function openExplorerLink (hash: string) {
```
juliangruber, line 14: It's not clear why this function needs to be curried, could you please remove the `return () =>` and use an arrow function in `WalletWidget` instead?

juliangruber (commented):
Great process! 👏

renderer/src/lib/utils.ts
davelange: It was to avoid having to use arrow functions every time it's used, but if it's not clear I'll remove it.

juliangruber (approved):
great work!

PR #1524, opened 2024-05-09
Add modules page

davelange:
Create unstyled modules page:
- Create mock modules data
- Create hook to fetch and return data


juliangruber (approved):
great!

PR #1544, opened 2024-05-21
Add chart controls

davelange:
- In the chart, adjust the scheduled rewards line to be relative to the total rewards received line.
- Add chart controls to filter historical data by time range.
- Improve the mock data generation, to have more data and more similar to the real data.


juliangruber (approved):
👏

PR #1549, opened 2024-05-22
Add restart `core`. Closes #1548

### LINKED ISSUE #1548: Automatically restart Station Core

juliangruber:
Closes #1548

### DIFF
--- a/main/core.js
+++ b/main/core.js
@@ -13,6 +13,7 @@ const { Activities } = require('./activities')
 const { Logs } = require('./logs')
 const split2 = require('split2')
 const { parseEther } = require('ethers/lib/utils')
+const { once } = require('node:events')
 
 /** @typedef {import('./typings').Context} Context */
 
@@ -43,7 +44,9 @@ async function setup (ctx) {
     }
   }
   await maybeMigrateFiles()
-  await start(ctx)
+  while (true) {
+    await start(ctx)
+  }
 }
 
 /**
@@ -63,6 +66,7 @@ async function start (ctx) {
     },
     stdio: ['pipe', 'pipe', 'pipe', 'ipc']
   })
+  console.log('Core pid', childProcess.pid)
 
   assert(childProcess.stdout)
   childProcess.stdout.setEncoding('utf8')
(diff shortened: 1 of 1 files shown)


juliangruber:
> This looks like a disruptive change to me. Did you check that the caller of `setup` can handle the case when `setup` never returns?

> 
> IIRC:
> 
> * now: `setup` returns immediately after starting Station Core process, the process runs in the background
> * your change: `setup` waits for the process to complete
> 
> If we decide this is the direction to go, then I think we should rename `start` to something like `run` or `runOnce`, considering that the function no longer starts the process in the background but instead waits until the process exits.

Yes, `core.setup()` is the last call performed by `run()` in `main/index.js`. We could change its name as to not break consistency with the other functions called `setup()`, what do you think about `start()`? `run()` also works. `runOnce` doesn't work, as the function restarts Core inside its body.

juliangruber:
> This looks like a disruptive change to me.

Are you saying we should find a different way to restart it?

bajtos:
> Yes, `core.setup()` is the last call performed by `run()` in `main/index.js`. We could change its name as to not break consistency with the other functions called `setup()`, what do you think about `start()`? `run()` also works. `runOnce` doesn't work, as the function restarts Core inside its body.

Thank you for checking!

Based on what you wrote, I think it will be best to rename `setup` to `run`.

> > This looks like a disruptive change to me.
> 
> Are you saying we should find a different way to restart it?

I'd like to be careful and make sure we understand the ramifications of this change.

It looks "easy" and innocent at first sight, but changes the behaviour in subtle ways that can easily introduce bugs - at least, that's what I remember from how we were iterating on `lib/zinnia.js` in Station Core.

main/core.js
```diff
@@ -122,34 +126,46 @@ async function start (ctx) {
...
+        ? `via signal ${exitSignal}`
+        : `with code: ${code}`
+      const msg = `Core exited ${reason}`
+      console.log(msg)
+      exitReason = exitSignal || code ? reason : null
+    })(),
+    (async () => {
+      const [code] = await once(childProcess, 'close')
+      controller.abort()
+      console.log(`Core closed all stdio with code ${code ?? '<no code>'}`)
```
bajtos, line 152: If I remember correctly my assumptions behind the original implementation: We rely on the fact that `exit` is emitted first and `close` is emitted second. That allows us to capture `exitReason` in the `exit` event handler and later use it in the `close` handler.

If that assumption is correct, then we don't need `Promise.all` and can simplify the code to listen for these events sequentially.

```js
const [code, exitSignal] = await once(childProcess, 'exit', { signal })
// ...
const [closeCode] = await once(childProcess, 'close')
// ...
```

WDYT?

main/core.js
```diff
@@ -43,7 +44,9 @@ async function setup (ctx) {
     }
   }
   await maybeMigrateFiles()
-  await start(ctx)
+  while (true) {
+    await start(ctx)
+  }
```
bajtos, line 49: I have a proposal to consider:

Let's keep the current `setup` semantics that's consistent with other Station Desktop backend components.

Change the `setup` function to return `run` function, change the code calling `setup` to collect all `run` functions and execute them as the last step.

```diff
- await start(ctx)
+ return run
```

Later in `index.js`:

```diff
- await core.setup(ctx)
+ const runCore = await core.setup(ctx)
+ await runCore(ctx)
```

This design makes it easy to add more long-running tasks in the future.

OTOH, maybe we don't need it yet and it's enough to rename `setup` to `run` 🤷🏻 

```diff
- await core.setup(ctx)
+ await core.run(ctx)
```

juliangruber:
> > Yes, `core.setup()` is the last call performed by `run()` in `main/index.js`. We could change its name as to not break consistency with the other functions called `setup()`, what do you think about `start()`? `run()` also works. `runOnce` doesn't work, as the function restarts Core inside its body.
> 
> Thank you for checking!
> 
> Based on what you wrote, I think it will be best to rename `setup` to `run`.

I added `run` in addition to `setup`, so that run doesn't modify `ctx`, which typically only the `setup` methods do.

> > > This looks like a disruptive change to me.
> > 
> > 
> > Are you saying we should find a different way to restart it?
> 
> I'd like to be careful and make sure we understand the ramifications of this change.
> 
> It looks "easy" and innocent at first sight, but changes the behaviour in subtle ways that can easily introduce bugs - at least, that's what I remember from how we were iterating on `lib/zinnia.js` in Station Core.

I agree, that migration was quite tricky!

main/core.js
juliangruber: I tried it like this, then force-quit the core process, and only this got logged:
 
```
Core exited via signal SIGKILL
```

And if I move the `close` handler first, only that gets logged.

I can't tell you exactly why, but unfortunately this approach doesn't work.

main/core.js
juliangruber: After https://github.com/filecoin-station/desktop/pull/1549/commits/896b93778b11b0b7c4a056ec0d69a4be718b2e64, would you still propose this?

main/index.js
```diff
@@ -174,6 +174,8 @@ async function run () {
     await wallet.setup(ctx)
     await telemetry.setup()
     await core.setup(ctx)
+
+    await core.run(ctx)
```
bajtos, line 178: 👏🏻

main/core.js
bajtos: Oh, maybe we added the other event listener too late after the event was already emitted.

```js
const onceExited = once(childProcess, 'exit', { signal })
const onceClosed = once(childProcess, 'close')

const [code, exitSignal] = await onceExited
// ...
const [closeCode] = await onceClosed
// ...
```

At this point, I am not sure if this is an improvement over your `Promise.all` solution. WDYT?

main/core.js
bajtos: I love https://github.com/filecoin-station/desktop/commit/896b93778b11b0b7c4a056ec0d69a4be718b2e64 😍

main/core.js
juliangruber: No that works, I think I prefer it :)

juliangruber: https://github.com/filecoin-station/desktop/pull/1549/commits/aeb6f82f637b9b0d89533ae3257e7646a92c663e

main/core.js
```diff
@@ -119,36 +129,36 @@ async function start (ctx) {
...
+  const exitReason = exitSignal || exitCode ? reason : null
+
+  const [closeCode] = await onceClosed
+  console.log(`Core closed all stdio with code ${closeCode ?? '<no code>'}`)
+
+  if (closeCode === 2) {
+    // FIL_WALLET_ADDRESS did not pass our screening. There is not much
+    // we can do about that, there is no point in reporting this error
+    // to Sentry.
+    throw new Error('Don\'t restart')
```
bajtos, line 154: What happens when this error is thrown? Is this going to crash the entire Station Desktop app? I guess that's acceptable, considering that this condition should never happen.

However, I think it will be confusing when a user is presented with "uncaught error" window saying "don't restart".

Let's rework this error message to better explain to the reader what happened.

bajtos (changes requested):

main/core.js
juliangruber: Like this?

bajtos (approved):
👌🏻

PR #1551, opened 2024-05-22
Update design system

davelange:
- Update the tailwind config. Add missing values and remove some unused styles. (There's more cleaning up to do, but it would be better to do that later)
 - Update fonts imported
 - Add a `Text` generic component. It's polymorphic, so we can define what html tag is used, and we can set its size, font, color and transform by props: `<Text size="sm" font="title" color="primary">Hello</Text>`
 - Add a `Button` generic. Also polymorphic, so it can be either a html button or a link. `<Button type="button" variant"primary" icon={<Icon />}>Click me</Button>`
 - Add a switch toggle component


juliangruber (approved):
👏

PR #1568, opened 2024-05-24
Style wallet page info

davelange:
Style info section of the wallet page
 - Add Tooltip component
 - Add Address component
 - Add new icons
 
Screenshot 2024-05-24 at 17 32 57


renderer/src/pages/wallet/Wallet.tsx
```diff
@@ -1,48 +1,58 @@
...
-                </div>
-            </section>
-
-            <section className="w-1/3">
-                <TransferForm />
-            </section>
-
-        </div>
+    <div className='w-full flex'>
+      <PageShell>
```
juliangruber, line 20: What's the purpose of the `PageShell` component? Could it be part of `Layout`? And why does it not contain the `w-full flex` element? Just for my understanding 🙏

renderer/src/components/Address.tsx
```diff
@@ -0,0 +1,60 @@
...
+import CopyIcon from 'src/assets/img/icons/copy.svg?react'
+import Tooltip from './Tooltip'
+import { useRef, useState } from 'react'
+
+const Address = ({
+  address
+}: {
+  address: string;
+}) => {
+  const [copyText, setCopyText] = useState('Copy')
```
juliangruber, line 12: instead of a string state, what about a boolean `isTextCopied`?

renderer/src/components/Address.tsx
```diff
@@ -0,0 +1,60 @@
...
+    navigator.clipboard.writeText(address)
+    setCopyText('Copied')
+
+    setTimeout(() => {
+      triggerRef.current?.blur()
+    }, 700)
+
+    setTimeout(() => {
+      setCopyText('Copy')
+    }, 800)
```
juliangruber, line 27:
```suggestion
    const timeouts = [
      setTimeout(() => {
        triggerRef.current?.blur()
      }, 700),
      setTimeout(() => {
        setCopyText('Copy')
      }, 800)
    ]
    return () => {
      for (const timeout of timeouts) {
        clearTimeout(timeout)
      }
    }
```

juliangruber (changes requested):

renderer/src/pages/wallet/Wallet.tsx
davelange: Different pages of the app have different layouts (Settings content has a centered max width, Dashboard has no max width, Wallet has no max width and has that extra section on the right), but share areas with the same padding/margin. PageShell provides that uniform padding/margin. It could be part of `Layout`, if it weren't for the section on the right in the modules page, so in this case it's easier for each page to compose its own layout with `PageShell`

But `PageShell` is probably not a good name, because it implies it wraps everything. I could change it to `Container`. wdyt?

juliangruber: Got it! Some suggestions:

- `Main` - since semantically it's also the `<main>` element
- `Container` - as you suggested, this is a well enough understood concept

> It could be part of Layout, if it weren't for the section on the right in the modules page

I don't want to bike shed this, but one idea: What about instead adding this to layout, and adding an option there, for the modules page?

Definitely let's go with the simplest solution here.

davelange: > What about instead adding this to layout, and adding an option there, for the modules page?

That would add some complexity, as then each page would need to import Layout, meaning the common components inside Layout (Sidebar, DraggableArea, etc) would need to go somewhere else. 
Honestly I'm ok with just using regular CSS for now though, as there are only 4 pages, and it may be easier to figure out a good abstraction later on.

juliangruber: > Honestly I'm ok with just using regular CSS for now though, as there are only 4 pages, and it may be easier to figure out a good abstraction later on.

Sounds good!

juliangruber (approved):
Great work!

PR #1572, opened 2024-05-27
Style transfer flow

davelange:
Style transfer flow:
 - Add new icons
 - Add Input component, and update Button with error and disabled state
 - Add an `open` prop to `Tooltip`, which allows us to open it without any user interaction on the trigger
 - Add a `useAddressValidation` hook that handles address validation for both address forms


renderer/src/pages/wallet/ConfirmTransfer.tsx
```diff
@@ -0,0 +1,45 @@
...
+import Text from 'src/components/Text'
+import { Wallet } from 'src/hooks/StationWallet'
+import { formatFilValue } from 'src/lib/utils'
+
+const ConfirmTransfer = ({
+  walletBalance,
+  hide,
+  transferAllFundsToDestinationWallet
+}: {
+    walletBalance: Wallet['walletBalance'];
```
juliangruber, line 12: Could we also write `Wallet.walletBalance`?

renderer/src/pages/wallet/TransferWrapper.tsx
```diff
@@ -0,0 +1,73 @@
...
+}: {
+  walletBalance: Wallet['walletBalance'];
+  destinationFilAddress?: Wallet['destinationFilAddress'];
+  stationAddress?: Wallet['stationAddress'];
+  processingTransaction : Wallet['processingTransaction'];
+  editDestinationAddress:Wallet['editDestinationAddress'];
+  transferAllFundsToDestinationWallet: Wallet['transferAllFundsToDestinationWallet'];
+  dismissCurrentTransaction: Wallet['dismissCurrentTransaction'];
+}) => {
+  const [showConfirm, setShowConfirm] = useState(false)
```
juliangruber, line 28:
```suggestion
  const [isShowingConfirm, setIsShowingConfirm] = useState(false)
```

`showConfirm` sounds like a function name to me

juliangruber (changes requested):

renderer/src/pages/wallet/ConfirmTransfer.tsx
davelange: Sadly, no 
https://www.totaltypescript.com/why-you-cant-use-dot-notation-on-types

renderer/src/pages/wallet/BalanceControl.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const hasSufficientBalance = Number(walletBalance) >= sendThreshold
+
+    if (processingTransaction) {
+      setStatus('processing')
+    } else if (!processingTransaction && hasSufficientBalance) {
+      setBalanceStateAfterComplete(status, 'idle')
+    } else if (!processingTransaction && !hasSufficientBalance) {
+      setBalanceStateAfterComplete(status, 'accruing')
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
davelange: This useEffect sets `status`, and that should only change if:
 - there's a change in `processingTransaction`
 - there's a change in wallet balance

If I added `status` to the deps array, it would run in a loop. The linter complains because I use the `status` value inside the `useEffect`, but it shouldn't be a dependency in this case

renderer/src/pages/wallet/BalanceControl.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const [status, setStatus] = useState<'accruing' | 'idle' | 'processing' | 'complete'>('idle')
+
+  const setBalanceStateAfterComplete = useCallback((oldStatus: typeof status, newStatus: typeof status) => {
+    if (oldStatus === 'processing') {
+      setStatus('complete')
+      setTimeout(() => setStatus(newStatus), 2000)
+    } else {
+      setStatus(newStatus)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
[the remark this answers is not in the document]
davelange: Actually it's not needed here, I'll remove it

renderer/src/pages/wallet/BalanceControl.tsx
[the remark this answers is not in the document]
juliangruber: Interesting, if it works I'm happy 👍

davelange:
Added the changes discussed: 
 - showing a disabled transfer button while balance is not sufficient, with an info tooltip

renderer/src/hooks/useCurrentTransactionStatus.tsx
```diff
@@ -0,0 +1,34 @@
+import { useEffect, useRef, useState } from 'react'
+import { FILTransactionProcessing } from '../../../shared/typings'
+
+const useCurrentTransactionStatus = (transaction?: FILTransactionProcessing) => {
+  const [status, setStatus] = useState<'none' | 'processing' | 'complete'>('none')
```
juliangruber, line 5: I should have brought this up earlier, but this type conflicts with that one, which we're already using throughout the codebase:

```ts
export type FILTransactionStatus = 'succeeded' | 'processing' | 'failed'
```

Could we please unify these types? `complete -> succeeded` is an obvious match, what about `none` though?

renderer/src/hooks/useCurrentTransactionStatus.tsx
```diff
@@ -0,0 +1,34 @@
...
+      timeout.current = setTimeout(() => {
+        setStatus('none')
+        setCurrentTransaction(undefined)
+      }, 1000)
+    }
+
+    return () => {
+      clearTimeout(timeout.current)
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
juliangruber, line 25: could we remove use of this exemption by converting `status` into a ref? I think disabling this rule is dangerous, because for example in a next PR we could add another variable that _should_ be listed, and eslint wouldn't tell us about it.

juliangruber (changes requested):

renderer/src/hooks/useCurrentTransactionStatus.tsx
davelange: I need a status for when there is no transaction. Maybe `idle` is better?

juliangruber: Why would a transaction have no transaction? I don't understand this state. As soon as you create a transaction it's processing.

juliangruber: what about returning `FILTransactionStatus || null`?

renderer/src/hooks/useCurrentTransactionStatus.tsx
juliangruber: This sounds great to me, it's also suggested on SO here: https://stackoverflow.com/questions/63224151/how-can-i-access-state-in-an-useeffect-without-re-firing-the-useeffect

It would be great if we could go through all cases where exhaustive deps is disabled, and re-enable it through this method

juliangruber (changes requested):

renderer/src/hooks/useCurrentTransactionStatus.tsx
davelange: The point of the hook is to know if there is a transaction processing, and if there is, what state it's in. I was combining that into a single state, but if you prefer I could separate the two, and return 
```
{
 status: FILTransactionStatus
 transaction?: FILTransaction
}
```

davelange: Hadn't seen the latest comment, that sounds good to me

davelange: Simplified it a bit to use the tx status directly, instead of keeping separate state for it

juliangruber (approved):
😍

juliangruber:
After resolving conflicts we can merge :)

PR #1576, opened 2024-05-28
Update rewards mock data

davelange:
- Add module level data in rewards mock generation. Calculate combined values on the FE
- Add chart controls for active module 
- Updated structure is:
```
{
  timestamp: string;
  totalRewardsReceived: Record<moduleId, number>;
  totalScheduledRewards: Record<moduleId, number>;
}
```

### DIFF
--- a/renderer/src/hooks/StationRewards.tsx
+++ b/renderer/src/hooks/StationRewards.tsx
@@ -1,6 +1,5 @@
 import { useEffect, useMemo, useState } from 'react'
 import { getScheduledRewards } from 'src/lib/station-config'
-import { sumAllRewards } from 'src/lib/utils'
 import { mockData } from 'src/pages/dashboard/mockData'
 
 async function getHistoricalRewardsData () {
@@ -11,6 +10,10 @@ export type RewardsRecord = {
   timestamp: string;
   totalRewardsReceived: Record<string, number>;
   totalScheduledRewards: Record<string, number>;
+}
+
+export function sumAllRewards (data: RewardsRecord['totalRewardsReceived']) {
+  return Object.values(data).reduce((acc, val) => acc + val, 0)
 }
 
 const useStationRewards = () => {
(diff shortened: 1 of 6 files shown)


renderer/src/hooks/StationRewards.tsx
```diff
@@ -1,5 +1,6 @@
 import { useEffect, useMemo, useState } from 'react'
 import { getScheduledRewards } from 'src/lib/station-config'
+import { sumAllRewards } from 'src/lib/utils'
```
juliangruber, line 3: could you please move this function to this file? Since it's only used in one place, I would rather prevent `utils` from becoming a big utility bag, and have it here

renderer/src/pages/dashboard/Chart.tsx
```diff
@@ -8,6 +8,7 @@ import {
 } from 'chart.js'
 import { Line } from 'react-chartjs-2'
 import { RewardsRecord } from 'src/hooks/StationRewards'
+import { getRewardValue } from 'src/lib/utils'
```
juliangruber, line 11: could you please also move this function, since it's only used in this file here?

juliangruber (changes requested):

renderer/src/lib/utils.ts
```diff
@@ -1,6 +1,7 @@
 import { BigNumber, FilecoinNumber } from '@glif/filecoin-number'
 import { browseTransactionTracker } from './station-config'
 import { delegatedFromEthAddress, ethAddressFromDelegated, newFromString } from '@glif/filecoin-address'
+import { RewardsRecord } from 'src/hooks/StationRewards'
```
juliangruber, line 4:
```suggestion
```

juliangruber (changes requested):

renderer/src/lib/utils.ts
juliangruber: just a small linter fix

juliangruber (approved):

PR #1598, opened 2024-06-06
Fix wallet hook and sync

davelange:
Why:
 - The UI is still not displaying updated transactions in sync with the balance updates. After a transaction is complete `updateBalance()` is called, and transactions are then updated in a timeout, so there is a delay between the updates.
 - The `useWallet` hook has some issues where it runs a function in a loop while there is a processing transaction, due to dependencies in hooks.

How:
 - In `wallet.js`, after a transfer is complete, refresh all state, not just balance.
 - Update `useWallet` hook, to prevent dependency issues
 - The `useCurrentTransactionStatus` hook is unnecessary, so remove it. The tracking of the current transaction is already done in the `useWallet` hook


juliangruber (approved):
Makes sense 👍!

PR #1599, opened 2024-06-06
Add grid

davelange:
Add animated grid background to the transfer flow.
 - Uses HTML canvas and no additional deps
 - The grid has a warp config, that determines where the warping is centered around (x,y), how many lines are affected, and how much they are warped.
 - Generates an array of rows and col data. The n closest lines to the center of the warp will be deformed. The amount of warp depends on their distance to the warp centre, and is calculated using an easing function. The result of this is used to set control points to render bezier curves.
 - Has a tween render method, that allows to animate the grid from one state to another over a given duration.
 - Also renders a midline and some target circles.
 - The grid config depends on the user balance. The higher the balance, the more warp.

Demo:
https://www.loom.com/share/0d09047c9d6040adbb89d3354a23058d


renderer/src/lib/grid.ts
```diff
@@ -0,0 +1,320 @@
...
+    this.ctx.strokeStyle = '#FFFFFF66'
+    this.calcGridLines()
+    this.renderDimension(this.rows, 'row')
+    this.renderDimension(this.cols, 'col')
+  }
+
+  renderDimension (line: Line[], dimension: 'row' | 'col') {
+    const isRow = dimension === 'row'
+
+    line.forEach((item) => {
```
juliangruber, line 218:
```suggestion
    for (const item of line) {
```

renderer/src/lib/grid.ts
```diff
@@ -0,0 +1,320 @@
...
+          isRow ? item.y2 : config.bendOutEnd
+        )
+        // line
+        this.ctx.lineTo(item.x2, item.y2)
+      } else {
+        this.ctx.lineTo(item.x2, item.y2)
+      }
+
+      this.ctx.stroke()
+    })
```
juliangruber, line 259:
```suggestion
    }
```

renderer/src/pages/wallet/GridCanvas.tsx
```diff
@@ -0,0 +1,74 @@
+import { useEffect, useRef } from 'react'
+import { Wallet } from 'src/hooks/StationWallet'
+import { Grid } from 'src/lib/grid'
+import { FILTransactionProcessing } from '../../../../shared/typings'
```
juliangruber, line 4: can this import path be refactored?

juliangruber (changes requested):
Great work!

renderer/src/pages/wallet/GridCanvas.tsx
davelange: Yes, I think it works id I add  a `"shared/*": ["../shared/*"]` entry to tsconfig, then it could be `'shared/typings'`

juliangruber (approved):
nice work!!!

PR #1618, opened 2024-06-13
Add read scheduled rewards from API

juliangruber:


renderer/src/hooks/StationRewards.tsx
```diff
@@ -1,9 +1,32 @@
...
-  return mockData
+type StatsResponse = {
+  day: string;
+  scheduled_rewards: string;
+}[]
+
+async function getHistoricalRewardsData (address: string) {
+  const res = await fetch(
+    'https://stats.filspark.com/participants/scheduled-rewards/daily' +
+      `?address=${address}&from=2024-06-13&to=2024-06-13`
```
davelange, line 13: Won't this always get just one single day?

juliangruber: You're right! I changed it to set dynamic from/to based on the mock implementation (last 100 days). This isn't enough though, shall we make this dependent on the graph timespan selected, or always fetch all days?

davelange: I think for now it makes sense to get all days. If that query gets too heavy, we could do an initial query only for 7 days, but that's not a problem for now.

PR #1629, opened 2024-06-18
Fixes 

davelange:
- Fix chart tooltip zindex
 - Force chart to have 0 minimum, as it was showing negative values
 - Disable spellcheck on address inputs


juliangruber (approved):
nice!

PR #1640, opened 2024-06-24
Fix link

davelange:
Why:
 - The link to the wallet docs is opening the link as an embed in the app, instead of opening the link in a browser.

How:
 - Add new record to safelist of links
 - Open link in browser

### DIFF
--- a/main/utils.js
+++ b/main/utils.js
@@ -25,8 +25,8 @@ const allowedURLs = [
 ].map(str => new URL(str))
 
 const allowedURLsRegEx = [
-  /https:\/\/beryx.zondax.ch\/v1\/search\/fil\/mainnet\/address\/.*$/,
-  /https:\/\/docs.filstation.app\/.*$/
+  /https:\/\/docs.filstation.app\/.*$/,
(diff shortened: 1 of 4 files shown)


main/utils.js
```diff
@@ -25,7 +25,8 @@ const allowedURLs = [
 ].map(str => new URL(str))
 
 const allowedURLsRegEx = [
-  /https:\/\/beryx.zondax.ch\/v1\/search\/fil\/mainnet\/address\/.*$/
+  /https:\/\/beryx.zondax.ch\/v1\/search\/fil\/mainnet\/address\/.*$/,
+  /https:\/\/docs.filstation.app\/.*$/
```
bajtos, line 29: Nitpick: let's add the new line to the top of the list, so that we don't need to change the existing code.

It's unfortunate that https://standardjs.com disallows trailing commas, but that's a different story.

```suggestion
  /https:\/\/docs.filstation.app\/.*$/,
  /https:\/\/beryx.zondax.ch\/v1\/search\/fil\/mainnet\/address\/.*$/
```

bajtos (approved):
Looks pretty good to me! I have one minor comment to consider.

bajtos:
@juliangruber hindsight please 🙏🏻

PR #1649, opened 2024-06-26
fix: rotate Sentry token

bajtos:
Old Station versions are producing too many un-actionable issues. We need to cut them off.

Related:
- #1606-


bajtos:
@juliangruber FYI.

I also learned that we can rate-limit individual Sentry tokens. I wish I knew about this feature before! I configured the rate limit for the legacy token to 1000 req/day.

https://space-meridian.sentry.io/settings/projects/filecoin-station/keys/ff9615d8516545158e186d863a06a0f1/

Screenshot 2024-06-26 at 09 34 30

PR #1673, opened 2024-07-04
Fix deselect time range

### LINKED ISSUE #1671: It's possible to deselect time range on the main screen
In the chart on the main screen, it’s possible to deselect time range, i.e. choose neither of 7d/1m/1y/all intervals. The chart does not show any data in such case.

Steps to reproduce: click twice on the same time range button.

Screenshot 2024-07-04 at 09 03 30

davelange:
- Disable time range button when it is active

Fixes #1671


juliangruber (approved):
awesome!

PR #1741, opened 2024-07-31
Chart: Fix tooltip precision by using bigint in more places

juliangruber:
Screenshot 2024-07-31 at 10 53 31


renderer/src/hooks/StationRewards.tsx
```diff
@@ -1,6 +1,7 @@
 import { useEffect, useMemo, useState } from 'react'
 import { getScheduledRewards } from 'src/lib/station-config'
 import useWallet from 'src/hooks/StationWallet'
+import { formattedFilToBigInt } from 'src/lib/utils'
```
juliangruber, line 4: eventually we should refactor the system never to pass formatted FIL around, only bigint or strings

bajtos:
Nice! 👏🏻

PR #1742, opened 2024-07-31
Add import seed phrase. Closes #1727

### LINKED ISSUE #1727: Add import seed phrase

juliangruber:
Closes #1727

Screenshot 2024-07-31 at 14 07 29

### DIFF
--- a/main/settings.js
+++ b/main/settings.js
@@ -31,6 +31,19 @@ async function setup (ctx) {
       clipboard.writeText(await wallet.getSeedPhrase())
     }
   }
+
+  ctx.importSeedPhrase = async () => {
+    const button = showDialogSync({
+      title: 'Import Seed Phrase',
+      // eslint-disable-next-line max-len
+      message: 'The seed phrase is used in order to back up your wallet, or move it to a different machine. Please copy it to your clipboard before proceeding. Please be cautious, as this will overwrite the seed phrase currently used, which will be permanently lost (unless backed up before).',
+      type: 'info',
+      buttons: ['Cancel', 'Import from Clipboard']
+    })
+    if (button === 1) {
+      await wallet.setSeedPhrase(clipboard.readText())
+    }
+  }
 }
 
 module.exports = {
--- a/main/settings.js
+++ b/main/settings.js
@@ -36,9 +36,9 @@ async function setup (ctx) {
     const button = showDialogSync({
       title: 'Import Seed Phrase',
       // eslint-disable-next-line max-len
-      message: 'The seed phrase is used in order to back up your wallet, or move it to a different machine. Please copy it to your clipboard before proceeding. Please be cautious, as this will overwrite the seed phrase currently used, which will be permanently lost (unless backed up before).',
+      message: 'The seed phrase is used in order to back up your wallet, or move it to a different machine. Please copy it to your clipboard before proceeding. Please be cautious, as this will overwrite the seed phrase currently used, which will be permanently lost.',
       type: 'info',
-      buttons: ['Cancel', 'Import from Clipboard']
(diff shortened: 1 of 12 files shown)


main/settings.js
```diff
@@ -31,6 +31,19 @@ async function setup (ctx) {
       clipboard.writeText(await wallet.getSeedPhrase())
     }
   }
+
+  ctx.importSeedPhrase = async () => {
+    const button = showDialogSync({
+      title: 'Import Seed Phrase',
+      // eslint-disable-next-line max-len
+      message: 'The seed phrase is used in order to back up your wallet, or move it to a different machine. Please copy it to your clipboard before proceeding. Please be cautious, as this will overwrite the seed phrase currently used, which will be permanently lost.',
```
bajtos, line 39: I'd like @patrickwoodhead to review this message.

Do we need to explain what is the seed phrase? I would expect the user to understand that if they decide to import the seed phrase,

I think it's also best to put the call to action as the last sentence in the message.

Should we also tell the user that they may want to export the current seed phrase?

```suggestion
      message: 'Please be cautious. This action will overwrite the seed phrase currently used, which will be permanently lost. Please copy the seed phrase to import to your clipboard before proceeding.',
```

bajtos (approved):
LGTM. I did not test the new functionality, though.


I think a UI with a password input that does not show the secret phrase would provide a better UX than asking the user to copy the seed phrase to the clipboard. It's a small improvement that may not be worth the effort required to implement it.

main/settings.js
juliangruber: For reference, we're also not explaining what a seed phrase is in the export section.

juliangruber: In your suggestion it says `Please copy the seed phrase to import`, but I think it needs to be `Please copye the seed phrase to your clipboard, to import`

juliangruber: +1 to moving the CTA to the end

juliangruber:
> LGTM. I did not test the new functionality, though.
> 
> I think a UI with a password input that does not show the secret phrase would provide a better UX than asking the user to copy the seed phrase to the clipboard. It's a small improvement that may not be worth the effort required to implement it.

An input field will require styling, which we don't yet have for this element in this location. It will also require a form, a button, with states, etc, which we don't have here.

I agree that that would be nicer though.

bajtos (changes requested):
Hang on, this isn't going to work. When we change the wallet private key, we will derive a different passphrase for Station Core and it won't be able to decrypt the file storing the Station ID.

See https://github.com/filecoin-station/desktop/issues/1726

main/settings.js
```diff
@@ -31,6 +31,19 @@ async function setup (ctx) {
...
+      type: 'info',
+      buttons: ['Cancel', 'Import from Clipboard and Restart']
+    })
+    if (button === 1) {
+      await wallet.setSeedPhrase(clipboard.readText())
```
bajtos, line 44: What happens when the text in the clipboard is not a valid seed phrase? How is Station going to handle that?

juliangruber:
- [ ] validate seed phrase
- [ ] re-encode station id

main/settings.js
patrickwoodhead: Yeah I agree that we need to ask the user to be cautious as the first thing. Also, I'm a bit unclear what the CTA is asking the user to do. Is it asking them to copy their current seed phrase out, or copy their new one into the clipboard so that they can paste it into the input box?

juliangruber: The user needs to have their current seed phrase copied in their clipboard, from which we will then import it

bajtos:
> re-encode station id

See https://github.com/filecoin-station/desktop/issues/1726#issuecomment-2262594686

PR #1848, opened 2024-09-28
Fix constant gas estimate

juliangruber:
The fixed 0.01 FIL wasn't working any more.


juliangruber:
I need to figure out later why adding 50% to the gas estimate is required for it to be accepted.

juliangruber:
Likely the 50% is the difference between calling a basic contract method and calling the filforwarder logic

juliangruber:
I hope 50% will be sufficient, and we can tweak it later

PR #1929, opened 2024-11-21
ci: add retry e2e tests

juliangruber:
Trying out the two main retry actions here

Motivation: It's annoying to restart runs because the e2e test suite failed with some arbitrary machine / tooling timeout. Example: https://github.com/filecoin-station/desktop/actions/runs/11945473489/job/33298198746?pr=1928

https://github.com/marketplace/actions/retry-step

https://github.com/marketplace/actions/retry-action


.github/workflows/ci.yml
```diff
@@ -91,11 +91,17 @@ jobs:
...
+        with:
+          command: xvfb-run -a npm run test:e2e
+          max_attempts: 2
+          timeout_minutes: 10
       - name: Test end-to-end (Windows / macOS)
-        run: npm run test:e2e
         if: ${{ matrix.os != 'ubuntu-latest' }}
+        uses: Wandalen/wretry.action@v3.7.3
+        with:
+          command: npm run test:e2e
```
juliangruber, line 104: This one is more flexible (it also allows you to retry actions themselves, not just commands) and has sensible defaults. But, it has less stars

juliangruber:
Interesting: It failed twice here: https://github.com/filecoin-station/desktop/actions/runs/11949394698/job/33308751848?pr=1929

Options:
- Add a delay between runs
- Add more attempts
- Could it be a worker problem, and we need to restart the whole workflow?

bajtos (approved):
:shipit: 

I don't see any obvious problems. Let's see how this works in practice and iterate on further improvements.

PR #1943, opened 2024-12-02
formattedFilToBigInt: fix float to bigint error due to lack of precision

juliangruber:
Fixes:

Screenshot 2024-12-02 at 17 34 36

This .0005 comes from multiplying by 1e18

The impact of this error is that scheduled rewards get shown as 0, therefore I'm going to merge and release before review.

### DIFF
--- a/renderer/src/lib/utils.ts
+++ b/renderer/src/lib/utils.ts
@@ -15,7 +15,7 @@ export function formatFilValue (value = '') {
 }
 
 export function formattedFilToBigInt (value: string) {
-  return BigInt(Number(value) * 1e18)
+  return BigInt(Math.floor(Number(value) * 1e18))
 }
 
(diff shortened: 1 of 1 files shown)


bajtos (commented):
LGTM as a quick stop-gap fix.

However: 

Why are we formatting FIL to a floating-point number first and then converting this back to attoFIL? This is very likely to loose precision during the process.

Can we please use the original attoFIL value instead of formatting attoFIL to a FIL string and then converting it back.

juliangruber:
I agree! I took a stab at this refactor before, but it's too big in scope and imo not worth the time right now

juliangruber:
Good idea, added a comment in https://github.com/filecoin-station/desktop/issues/1183#issuecomment-2516671403

PR #2007, opened 2025-01-16
Rebrand banner in Desktop App. Closes #2004

### LINKED ISSUE #2004: Add Info Banner for Brand Update
As we mvoe from "Filecoin Station" to "Checker Node", we will need a banner that will keep operators informed about any upcoming changes, or what has just changed.

patrickwoodhead:
Closes #2004


renderer/src/components/RebrandBanner.tsx
```diff
@@ -0,0 +1,36 @@
+import Text from './Text'
+import { useState } from 'react'
+import CloseIcon from 'src/assets/img/icons/close.svg?react'
+import { openExternalURL } from 'src/lib/station-config'
+import { useLocation } from 'react-router'
+import { ROUTES } from 'src/lib/routes'
+
+const RebrandBanner = () => {
+  const { pathname } = useLocation()
+  const [visible, setVisible] = useState<boolean>(true)
```
juliangruber, line 10: This state isn't persisted, which means that on every restart of the app the banner will reappear. I think that should be ok, but wanted to double check if that's what you intended. If not, we can persist this state, which I can add to this PR if you want to

renderer/src/components/RebrandBanner.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  return (
+    <div className='absolute top-0 left-0 pl-28 pt-6 flex'>
+      <div className='bg-primary ml-1 p-6 pr-12 left-40 flex gap-5 rounded-lg'>
+        <Text font='mono' size='xs' bold className="text-white">🚨 Filecoin Station is becoming the Checker Network. <button onClick={() => openExternalURL('https://blog.checker.network/posts/why-web3-needs-the-checker-network')}><Text font='mono' size='xs' bold className="text-white underline" >Read more</Text></button></Text>
+        <button
+          className="absolute top-8 right-2"
+          onClick={() => setVisible(false)}
+        >
+          <CloseIcon />
```
juliangruber, line 28: Screenshot 2025-01-19 at 11 21 13

In the current light, I almost couldn't see the X. Can we change its color to something light, to improve the contrast?

juliangruber (changes requested):

renderer/src/components/RebrandBanner.tsx
patrickwoodhead: I'm happy for it to reappear after a refresh as a reminder. I decided to leave it this way intentionally, when I was thinking of whether I needed to do more. I don't think it will affect the user experience and it will remind the operator of upcoming changes. This banner and its message may easily change within the next few weeks too.

juliangruber (approved):

PR #2049, opened 2025-02-10
station core -> checker node

juliangruber:
For https://github.com/CheckerNetwork/desktop/issues/2048

### DIFF
--- a/build/after-pack.js
+++ b/build/after-pack.js
@@ -22,7 +22,7 @@ exports.default = async function (context) {
         : null
   if (arch === 'universal') return
   assert(arch, `Unknown architecture id: ${context.arch}`)
-  console.log(`Rebuild Station Core for arch=${arch}`)
+  console.log(`Rebuild Checker Node for arch=${arch}`)
   const ps = spawn(
     'node',
     ['scripts/post-install.js'],
@@ -32,7 +32,7 @@ exports.default = async function (context) {
         'Filecoin Station.app',
         'Contents',
         'Resources',
-        'core'
+        'checker'
       ),
       env: {
         ...process.env,
--- a/main/checker-node.js
+++ b/main/checker-node.js
@@ -16,16 +16,16 @@ const { parseEther } = require('ethers/lib/utils')
 const { once } = require('node:events')
 const { format } = require('node:util')
 
-const log = require('electron-log').scope('core')
+const log = require('electron-log').scope('checker')
 
 /** @typedef {import('./typings').Context} Context */
 
-// Core is installed separately from `node_modules`, since it needs a
+// Checker Node is installed separately from `node_modules`, since it needs a
 // self-contained dependency tree outside the asar archive.
-const corePath = app.isPackaged
-  ? join(process.resourcesPath, 'core', 'bin', 'station.js')
-  : join(__dirname, '..', 'core', 'bin', 'station.js')
-log.info(format('Core binary: %s', corePath))
+const checkerNodePath = app.isPackaged
+  ? join(process.resourcesPath, 'checker', 'bin', 'checker.js')
+  : join(__dirname, '..', 'checker', 'bin', 'checker.js')
+log.info(format('Checker Node binary: %s', checkerNodePath))
 
 const logs = new Logs()
 const activities = new Activities()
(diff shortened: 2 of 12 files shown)


build/after-pack.js
```diff
@@ -32,7 +32,7 @@ exports.default = async function (context) {
         'Filecoin Station.app',
         'Contents',
         'Resources',
-        'core'
+        'checker'
```
bajtos, line 35: Since Filecoin Station will be eventually called Checker App or simply Checker, I think it would be better to call this directory `checker-node`. 

Such a name is also more consistent with the filename `main/checker-node.js` I see later in this pull request.

WDYT?

main/checker-node.js
```diff
@@ -16,16 +16,16 @@ const { parseEther } = require('ethers/lib/utils')
 const { once } = require('node:events')
 const { format } = require('node:util')
 
-const log = require('electron-log').scope('core')
+const log = require('electron-log').scope('checker')
```
bajtos, line 19: Since you rename the file from "core" to "checker-node", I think the logging scope should follow the same pattern.

```suggestion
const log = require('electron-log').scope('checker-node')
```

Thoughts?

main/checker-node.js
```diff
@@ -76,24 +76,24 @@ function maybeReportErrorToSentry (err, scopeFn) {
  * @param {Context} ctx
  */
 async function start (ctx) {
-  log.info('Starting Core...')
+  log.info('Starting Checker...')
```
bajtos, line 79: For consistency with messages printed from other places:

```suggestion
  log.info('Starting Checker Node...')
```

main/checker-node.js
```diff
@@ -76,24 +76,24 @@ function maybeReportErrorToSentry (err, scopeFn) {
...
-    corePath,
-    ['--json', '--recreateStationIdOnError'],
+    checkerNodePath,
+    ['--json', '--recreateCheckerIdOnError'],
     {
       env: {
         ...process.env,
         FIL_WALLET_ADDRESS: await wallet.getAddress(),
-        PASSPHRASE: await wallet.signMessage('station core passhprase'),
+        PASSPHRASE: await wallet.signMessage('checker node passhprase'),
```
bajtos, line 88: This is a breaking change. The Checker Node won't be able to decrypt the Station ID file encoded using the old passphrase derived from "station core passhphrase". Since we are using `--recreateCheckerIdOnError`, the Checker Node will create a new checker id.

Is that your intent?

While the upgrade is rolled to the network, we will see twice as many Checker IDs in our telemetry.

Are we okay with that?

My suggestion is to keep the old passphrase for backward compatibility.

```suggestion
        // Note: we must keep the passphrase to avoid rotation of Checker IDs
        PASSPHRASE: await wallet.signMessage('station core passhprase'),
```

OTOH, if we decide to change the passphrase, then let's choose a new value that we won't need to change in the future.

```suggestion
        PASSPHRASE: await wallet.signMessage('passhprase seed'),
```

bajtos (changes requested):
Great work!

My main concern is about preserving backwards compatibility.

1. See the comment below regarding PASSPHRASE.

2. IIUC, this pull request does not change the desktop app name and the location of STATE_ROOT/CACHE_ROOT directories, therefore upgrade to this new version will not affect things like the job counter stored in STATE_ROOT. Can you please confirm?

build/after-pack.js
bajtos: If you agree with my proposal, then please remember to update other files too, e.g. `.eslintrc.json`, `.gitignore` and `electron-builder.yml`.

juliangruber: > Since Filecoin Station will be eventually called Checker App or simply Checker

For correctness: "Filecoin Station" will not be renamed to Checker, it is becoming "Checker Network". "Filecoin Station Desktop" is becoming "Checker App" or short "Checker".

juliangruber: a301a49

main/checker-node.js
juliangruber: Done

main/checker-node.js
juliangruber: a301a49

main/checker-node.js
juliangruber: My thinking was that a one time rotation of all ids shouldn't be a problem, but I agree that simply keeping the old pass phrase makes reporting simpler, while changing the passphrase doesn't add any value. I'll revert that part

juliangruber: a48603a

bajtos (approved):
The code changes LGTM.

Can you please respond to my previous comment?

> IIUC, this pull request does not change the desktop app name and the location of STATE_ROOT/CACHE_ROOT directories, therefore upgrade to this new version will not affect things like the job counter stored in STATE_ROOT. Can you please confirm?

juliangruber:
> The code changes LGTM.
> 
> Can you please respond to my previous comment?
> 
> > IIUC, this pull request does not change the desktop app name and the location of STATE_ROOT/CACHE_ROOT directories, therefore upgrade to this new version will not affect things like the job counter stored in STATE_ROOT. Can you please confirm?

Sorry, I missed this.

Yes, this is not affected, I verified this manually.

